PR #12, opened 2022-01-10
Fix Gitpod support

### LINKED ISSUE #11: Gitpod support isn't fully operational
@RCheesley I tried to run gitpod and preview the pages. i had these errors

Screenshot 2022-01-10 at 00 21 49

_Originally posted by @fakela in https://github.com/mautic/user-documentation/issues/9#issuecomment-1008445203_

RCheesley:
This fixes #11 by:

- Removing some files which seemed to be from a demo install
- Moving the content into the /docs folder rather than having another sub-folder /source
- Fixing incorrect paths to the requirements.txt file
- Adding a couple of settings which were missing from the config file


RCheesley:
OK I think we are up and running now.

In this PR I have:

- Removed some files which seemed to be from a demo install
- Moved the content into the /docs folder rather than having another sub-folder
- Fix incorrect paths to the requirements.txt file
- Added a couple of settings which I had missed from the config file 

One thing I could not get working which we had in the dev docs was the extension 'xref' which I believe relates to the [external links extension](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-the-external-links-extension) but I can't see any reference to it in up to date docs, so it may be a relic from an old tutorial. Will need to test to see if that still works.

Testing link: https://gitpod.io/#https://github.com/mautic/user-documentation/pull/12

Note: Pages need to be added to the TOC to show in the nav bar :)

RCheesley:
Note: Please make sure you do select to install with these:
Welcome to GitLens — user-documentation — Gitpod Code (1)

PR #26, opened 2022-02-20
Added How to update Mautic

favour-chibueze:


fakela (commented):
@favour-chibueze what do you think should we add the links normally lie `[Hello](https://github.com/mautic/user-documentation/pull/26#pullrequestreview-905918536)` or `[Hello] 
 (hello)`

fakela:
> @favour-chibueze what do you think should we add the links normally like this `[Hello](https://github.com/mautic/user-documentation/pull/26#pullrequestreview-905918536)` or `[Hello](hello-link)`

@RCheesley

RCheesley:
Actually neither is correct for the new docs.

Here is the documentation on the external link extension we should be using with Read the Docs:

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-the-external-links-extension

We've got it set up and working great already on the dev docs in case you need a reference point: https://github.com/mautic/developer-documentation-new/tree/main/docs/links

fakela:
> Actually neither is correct for the new docs.
> 
> Here is the documentation on the external link extension we should be using with Read the Docs:
> 
> https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-the-external-links-extension
> 
> We've got it set up and working great already on the dev docs in case you need a reference point: https://github.com/mautic/developer-documentation-new/tree/main/docs/links

Does the new pattern work if we are using Markdown in the docs and not RST

PR #42, opened 2022-04-07
Add mautic 3 upgrade docs (#35)

fakela:


RCheesley:
@fakela is there a reason why we need this additional PR, or did I miss something with #35?

PR #63, opened 2022-04-28
Updated companies with correct rst format

Kelvindest:


RCheesley:
Hi @Kelvindest there were a lot of problems that Vale was reporting with this PR which should be fixed before it was merged. I've reverted the PR accordingly, can you please make the PR again, and we'll review the things which need to be fixed, and merge once both the GitHub Actions are passing.

Thanks, and sorry we didn't pick this up before we merged the PR.

PR #97, opened 2022-08-17
TEDU-107 added segment section

favour-chibueze:
TEDU-107 segment section


docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+This accomplishes the same thing as creating a Dynamic Segment with a filter for Points, the operator ``greater than (or equal to)``, and the minimum number of Points. The difference is if you only want to add Contacts to a Segment who reach the point value after creating this trigger to a Segment, you can.
+
+.. vale off
+
+CSV upload
+===========
+
+.. vale on
+
+If you have a list outside of Mautic, saved as a UTF-8 encoded CSV file, you can upload the list directly into a Mautic Segment.
```
RCheesley, line 222: Maybe link here to the import section in managing contacts (may not be written yet!) - wonder if this documentation should sit there and we just link to it? What do you think? 

There's stuff like using the import in the background cron which are not relevant to add here (we just want to focus on the bit about how you get them into a segment), so I think it makes sense to point the user to the more in-depth docs.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+1. Create a new Segment by clicking the **+New** button.
+
+2. In the **Details** tab, add a **Name**, **Public name** and **Description** to your Segment.
+
+3. Click the **Filters** tab to add filters.
+
+4. Click the **Choose one**… menu and search for the field you’d like to Segment by. Listed below are three types of fields:
+      
+  * Contact fields
```
RCheesley, line 265: Same issue here with the indentation - maybe it is 4 rather than 3 spaces?

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+    * Set Fields to **Available for Segments = Yes** in your Custom Field manager to appear here.
+    * Contacts associates with multiple Companies, but Mautic adds them to Segments based on fields for the primary Company.
+
+5. Add more filters, using the **And** and **Or** operators. An **Or** operator creates a new group of filters which can include And operators.
+
+6. Click **Save and close**.
+
+.. note:: 
+
+    The following filters rebuild immediately in real time, while other filters rebuild with the nightly refresh:
```
RCheesley, line 279: This is Acquia-specific so we should instead be replacing this with a note that segments are rebuilt according to how frequently you fire your cron jobs, and pointing people at the cron jobs docs about building segments.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+    * All core fields
+    * Custom fields added by the User
+    * Category
+    * Tags
+    * Contact Email received
+    * Hit URL
+    * Owner ID
+
+    * The nightly Segment rebuild jobs for each Segment are randomized for six hours. Jobs are executed anytime between midnight to 6 AM with respect to the **Default timezone** set for the instance in **Settings > Configurations > System Settings**.
```
RCheesley, line 289: This is Acquia specific so needs to be removed. The two points below can remain.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+You must ensure that you use the correct operator and time frame to build an effective Segment.
+
+Once you’ve selected a date field as your filter, such as the default **Date last active** field or a custom **Birthday** field, you’ll have a list of operators to choose from:
+
+* **Equals** - The value on the Contact record exactly matches the filter value.
+
+* **Not Equal** - The value on the Contact record is any value that doesn’t match the filter value.
+
+* **Greater than** - The value on the Contact record is at a later date in time than X date. For example, **Greater than** today means anytime from tomorrow until the end of time.
```
RCheesley, line 324: In the front end it is a tiny bit confusing having the second bit of text in bold, especially when it lines up with the bold texts at the start of the line. I might suggest seeing how it looks using an inline code markup rather than bold, perhaps. See what you think!

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+* **Not like** - This operator isn't supported in date or date-time fields, and shouldn’t be used.
+
+* **Regexp** - Segment includes Contacts with values that match the specified regular expression pattern. If you aren't proficient with regular expression, don’t use this operator.
+
+* **Not regexp** - Segment includes Contacts with values that don’t match the specified regular expression pattern. If you aren't proficient with regular expressions, don’t use this operator.
+
+* **Starts with** - Segment includes Contacts whose field values begin with the specified numbers. These filter values should generally reference years (or years and months).
+
+For example, A value of 19 matches any Contacts whose field value has a year in the 1900&nbsp;s. A value of 200 matches Contacts with a year value between 2000 and 2009 and a value of 2020-11 matches Contacts with a field value in November 2020.
```
RCheesley, line 346: This looks a bit out of place as it has escaped the list. Please also fix the ascii characters.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+* **Ends with** - Segment includes Contacts whose field values end with the specified numbers. These filter values should generally reference days (or months and days).
+
+For example, a value of 1 matches anyone whose field value is on the 1&nbsp;st, 21&nbsp;st, or 31&nbsp;st of any month but a value of 01 matches the 1&nbsp;st of a month. A value of 01-01 finds Contacts whose value is for January 1 of any year.
```
RCheesley, line 350: This looks odd, and also you have some ascii chars which need to be fixed.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+* ``%My string`` is the same as ``ends with`` filter with ``My string`` value.
+
+* ``%My string%`` is the same as ``contains`` filter with ``My string`` value.
+
+* ``My string`` is the same as ``contains`` filter with ``My string`` value.
+
+A few notes for text filters:
+
+* ``starts with``, ``ends with``, ``contains`` filters were added later than the ``like`` one, so you can easily specify what you need now.
```
RCheesley, line 379:
```suggestion
* ``starts with``, ``ends with``, ``contains`` filters should be used rather than ``like`` as they are more specific, and therefore can be more performant.
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+*  ``%`` character in the middle of the string has no special meaning. ``contains`` filter with ``my % string`` will search for a string with ``%`` in the middle. The same is TRUE for ``like`` filter with ``%my % string%`` value. There is no need for escaping this character.
```
RCheesley, line 381:
```suggestion
*  A ``%`` character in the middle of the string has no special meaning. A ``contains`` filter with ``my % string`` will search for a string with ``%`` in the middle. The same is TRUE for a ``like`` filter with ``%my % string%`` value. There is no need to escape this character.
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+* Mautic searches for ``%`` character in a value for ``like`` filter and no modification is performed if at least 1 ``%`` is found.
```
RCheesley, line 383:
```suggestion
* Mautic searches for the ``%`` character in a value for the ``like`` filter and no modification is performed if at least one ``%`` is found.
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+.. image:: images/date-filters.png
+    :alt: Screenshot showing DatePicker.
+
+However, you can specify much more here. Mautic recognizes relative formats too (these string aren't translatable):
+
+* ``+1 day`` (you can also use ``1 day``)
+* ``-2 days`` (you can also use ``2 days ago``)
+* ``+1 week`` / ``-2 weeks`` / ``3 weeks ago``
+* ``+5 months`` / ``-6 months`` / ``7months ago``
```
RCheesley, line 410:
```suggestion
* ``+5 months`` / ``-6 months`` / ``7 months ago``
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+* ``Date identified equals last week`` returns all Contacts identified between 2022-03-01 and 2022-03-07 (Monday - Sunday).
+* ``Date identified less than last week`` returns all Contacts identified before 2022-02-22.
+* ``Date identified equals last month`` returns all Contacts identified between 2022-02-01 and 2022-02-28.
+* ``Date identified greater or equal last year`` returns all Contacts identified 2021-01-01 and after.
+* ``Date identified greater than last year`` returns all Contacts identified after 2021-12-31.
+* ``Custom Contact date field equal birthday -1 day`` returns all Contacts identified every year on 03-04 (4th march).
+* ``Custom Contact date field equal anniversary -1 month`` returns all Contacts identified every year on 02-04 (4th february)
+
+Once you have created your Segment, any applicable Contact automatically add through the execution of a cron job. This is the essence of Segments.
```
RCheesley, line 440:
```suggestion
Once you have created your Segment, any applicable Contact will be automatically added through the execution of a cron job. This is the essence of Segments.
```
Maybe consider linking to the relevant cron job documentation.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+To keep the Segments current, create a cron job that executes the following command at the desired interval:
+
+``php /path/to/mautic/bin/console mautic:Segments:update``
```
RCheesley, line 444:
```suggestion
``php /path/to/mautic/bin/console mautic:segments:update``
```

RCheesley: I think we should instead link to the docs, personally, rather than duplicate here.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+Through the execution of that command, it adds Contacts that match the filters and removes Contacts that no longer match. Any Contacts that were manually added will remain part of the list regardless of filters.
```
RCheesley, line 446:
```suggestion
Through the execution of that command, Mautic adds Contacts that match the filters and removes Contacts that no longer match. Any Contacts that were manually added will remain part of the list regardless of filters.
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+.. vale off
+
+Delete all Contacts in a Segment
+********************************
+
+.. vale on
+
+Filter the Contacts in the Segment. The batch delete action in the Contact table allows deletion of up to 100 Contacts at one time.
```
RCheesley, line 455: What is the purpose of having the 'Filter the Contacts in a Segment' here?

If it is an instruction (eg that they need to first create the segment) we should be clearer here.

Maybe:

'It is possible to delete all contacts in a segment manually rather than with a Campaign action. To do this, first create a segment with filters that will pick up all the Contacts you want to delete.'

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+.. vale off
+
+Delete all Contacts in a Segment
+********************************
+
+.. vale on
+
+Filter the Contacts in the Segment. The batch delete action in the Contact table allows deletion of up to 100 Contacts at one time. 
+This is a performance precaution since deleting more Contacts at one time could cause issues. Hundreds of Contacts can use this feature.
```
RCheesley, line 456:
```suggestion
This is a performance precaution since deleting more Contacts at one time could cause issues. You can, however, delete multiple batches of 100 contacts to delete larger lists.
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+1. Select the checkboxes next to those Contacts.
+
+2. Click the green arrow which appears at the top of the column.
+
+3. Select **Delete Selected** from the list.
+
+4. Choose the **Contacts** you’d like to delete in a Segment.
```
RCheesley, line 464: You already did this in step 1, I don't think it needs to be repeated here?

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+5. Click **Delete**.
+
+.. image:: images/select-delete.png
+    :width: 200
+    :alt: Screenshot showing Deleting all Contacts in a Segment.
+
+But deleting thousands of Contacts this way in one Segment becomes a tedious task. Luckily, there is a trick how to let the background workers do the job for you.
```
RCheesley, line 472:
```suggestion
Deleting thousands of Contacts this way in one Segment becomes a tedious task. Luckily, there is a trick how to let the background workers do the job for you.
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+1. Create a simple Campaign which has the Segment as the source
+
+2. Use the :ref:`Delete contact action<using the campaign builder>`.
+
+This way the ``mautic:campaign:update`` and ``mautic:campaign:trigger`` commands deletes all the Contacts in the Segment. As well as all the Contacts which adds to the Segment in the future.
```
RCheesley, line 478:
```suggestion
This way the ``mautic:campaign:update`` and ``mautic:campaign:trigger`` commands delete all the Contacts in the Segment, and all the Contacts added to the Segment in the future. 
```

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+But deleting thousands of Contacts this way in one Segment becomes a tedious task. Luckily, there is a trick how to let the background workers do the job for you.
+
+1. Create a simple Campaign which has the Segment as the source
+
+2. Use the :ref:`Delete contact action<using the campaign builder>`.
+
+This way the ``mautic:campaign:update`` and ``mautic:campaign:trigger`` commands deletes all the Contacts in the Segment. As well as all the Contacts which adds to the Segment in the future. 
+It's all done automatically in the background.
+It's necessary to configure the cron jobs. Deleted Contacts can't be recovered.
```
RCheesley, line 480: Link to the relevant section of the cron jobs documentation.

Personally I would put the part about the contacts not being recoverable in a big red warning box and advise folks to use this feature very carefully!

RCheesley (changes requested):
Great job on this rather long page!

A few edits and suggestions to improve! 🚀

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+.. vale on
+
+Once a Contact has accrued an assigned number of Points, the system can add them to a Segment. This may be a Segment for your most engaged Contacts who become eligible for special offers, or a Segment your sales team reviews to find strong prospects.
+
+.. image:: images/points-trigger.png
+    :alt: Screenshot showing Points trigger.
+
+1. In the **Points** section of the platform, select** Manage Triggers**.
+
+2. Click **+New**.
```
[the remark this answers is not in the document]
favour-chibueze: I agree since that's what we have in Mautic.

docs/segments/manage_segments.rst
```diff
@@ -3,4 +3,493 @@
...
+
+3. Give your trigger a **Name** and **Description**.
+
+4. Enter the **Minimum number of Points** for adding a Contact to your Segment.
+
+5. Decide if you’d like to add all Contacts with at least that number of Points to the Segment:
+ 
+  * If you only want to add new Contacts who reach the threshold to this Segment, select **No** (default).
```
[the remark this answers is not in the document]
favour-chibueze: Using 'tip' markup will be the best because I have tried fixing the indentation and it's still same

docs/segments/manage_segments.rst
favour-chibueze: I tried 4 spaces and it is still same

docs/segments/manage_segments.rst
favour-chibueze: yeah, this section is in the contacts documentation which has not been written yet

fakela (commented):
Looks good to me

favour-chibueze:
@RCheesley Please review this PR.

RCheesley (changes requested):
Great work on this PR @favour-chibueze! I have added some suggestions, one to apply across all the changes is checking the apostrophes are using the right format. It is flagged up by VSCode on Gitpod so I think you could copy the wonky one and do a find/replace hopefully!

I also re-opened a comment about linking to a section as it has been written now :)

.vscode/settings.json
```diff
@@ -4,6 +4,5 @@
     "vale.valeCLI.path": "/sphinx-tools/vale/vale",
     "esbonio.sphinx.buildDir": "${workspaceFolder}/build",
     "esbonio.sphinx.confDir": "${workspaceFolder}/docs",
-    "restructuredtext.pythonRecommendation.disabled": true,
-    "restructuredtext.preview.name": "sphinx"
```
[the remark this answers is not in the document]
favour-chibueze: no

PR #105, opened 2022-10-13
Add translation files

RCheesley:
This PR adds the necessary configuration file and the files for all our existing translations.

Generated following this workflow: https://docs.readthedocs.io/en/stable/guides/manage-translations-sphinx.html

To be maintained in sync with the GitHub bot from Transifex, hopefully, which requires the files to exist already.

The strings are pushed up to https://www.transifex.com/mautic/mautic-end-user-documentation/content/.

<a href="https://gitpod.io/#https://github.com/mautic/user-documentation/pull/105"></a>


RCheesley:
I am going to go ahead and merge this as it is needed to get the automation working, which is required to set up the automatic push/pull.

@favour-chibueze it is worth mentioning that I had to manually change the header on the search_operators.rst file as it was throwing an error. I am sure it will be tweaked when you get to that section anyway so I did not commit the change.

PR #133, opened 2022-12-15
Exclude links known to wrongly report broken

RCheesley:
This PR adds to the conf.py some links which are consistently causing our GitHub Actions to fail when they do actually work but they are being blocked or have anchors which confuses the GitHub Action.

It is used on the Dev Docs and I hadn't come across it before - it should be really useful!


RCheesley:
Wahooo! All green! @favour-chibueze @fakela this PR explicitly prevents the URLs we know are reporting as broken from being checked - therefore we know a green check = good to go and a red cross = a broken link somewhere.

Happened to spot it in the dev docs!

PR #136, opened 2023-01-13
Update campaign_builder.rst - Added Transactional No DNC text to M5 branch

jos0405:
Added Transactional No DNC text to M5 branch


docs/campaigns/campaign_builder.rst
```diff
@@ -79,6 +79,7 @@ The actions that Mautic offers in a Campaign include:
      - Sends a Webhook to a defined URL, using the GET, POST, PUT, PATCH, or DELETE methods. Headers and data is customizable, and support the use of tokens, such as Contact fields and the Contact's IP address. For example, ``{contactfield=firstname}``
    * - **Send Email**   
      - Sends a transaction or marketing Email to the selected Contact. You can send a transactional Email to the Contact multiple times. You can only send a marketing Email to the Contact  once across multiple sources. If the Contact has already received this Email from another source or the current Campaign, they aren't sent the Email again and the Contact progresses through the Campaign.
+     - Emails can be transactional or marketing. Marketing emails will obey Email Channel's Do Not Contact setting of the contact, Transactional Emails will disregard the Do Not Contact value. This allows to send Transactional emails as campaign step to Contacts even if they opted out to recieve Marketing emails via standard unsubscribe action and have DNC set as 'no'. Enable transactional email ignoring DNC can be turned on in Email Settings.
```
RCheesley, line 82:
```suggestion
     - Sends a transaction or marketing Email to the selected Contact. You can send a transactional Email to the Contact multiple times. You can only send a marketing Email to the Contact once across multiple sources. If the Contact has already received this Email from another source or the current Campaign, they aren't sent the Email again and the Contact progresses through the Campaign. Emails can be transactional or marketing. Marketing Emails obey Do Not Contact (DNC) settings of the Contact, Transactional Emails disregard the DNC value. This allows sending Transactional Emails as a Campaign step to Contacts, even if they opted out to receive marketing Emails via a standard unsubscribe action and therefore have DNC set for the Email Channel. Turned on by default in Mautic 5.0, it's possible to override the setting in Email Settings.
```

Just like we test code PR's in Gitpod, please review your docs edits in Gitpod too! This one broke the RST table layout, and there were several Vale fixes to do after addressing that 😸 .

There should be a Gitpod button to click and you'll see any issues in the 'output' tab.

As a side thought, this is a big wall of text to have in a small area in a table. Maybe it makes more sense to have a dedicated section which explains transactional and marketing emails, and link to it from this with a small description? WDYT @jos0405 ? I think a sub-section of emails.rst under Channels would make the most sense.

docs/campaigns/campaign_builder.rst
```diff
@@ -79,6 +79,7 @@ The actions that Mautic offers in a Campaign include:
      - Sends a Webhook to a defined URL, using the GET, POST, PUT, PATCH, or DELETE methods. Headers and data is customizable, and support the use of tokens, such as Contact fields and the Contact's IP address. For example, ``{contactfield=firstname}``
    * - **Send Email**   
      - Sends a transaction or marketing Email to the selected Contact. You can send a transactional Email to the Contact multiple times. You can only send a marketing Email to the Contact  once across multiple sources. If the Contact has already received this Email from another source or the current Campaign, they aren't sent the Email again and the Contact progresses through the Campaign.
```
RCheesley, line 81:
```suggestion
```

RCheesley (changes requested):
I think it would be best to have a section on the Channels > emails.rst file which explains Transactional v Marketing (and maybe also Personal emails and form results emails) with the key differences.

We could then link to it from within this table, making it less of a wall of text in a small table cell!

WDYT @jos0405 ?

RCheesley:
@jos0405 would you be up for refactoring this based on the feedback? I feel it's too much of a wall-of-text to go into a table as it stands.

adiux:
@jos0405 I think we should get this merged, as the related PR is also merged. Could you look into it?

adiux:
I'm actually a bit confused. Does it remove the unsubscribe header only from the transactional emails? Or also from the individual emails (sent from the contact details page), as it says in the PR description: 
> This PR removes an unsubscribe header from an individual and transactional emails.

From my tests it only removes it from the transactional emails.

docs/campaigns/campaign_builder.rst
adiati98: @RCheesley I'm not quite understand about the content. But from your responses here, can we link the KB article (https://kb.mautic.org/article/understand-the-difference-transactional-and-marketing-emails-in-mautic.html) here like so:

```suggestion
* - **Send Email**   
  - Sends a transactional or marketing Email to the selected Contact. Learn more about transactional and marketing Email in (this article).
```

adiati98: If that's okay, I can help to apply it.

adiati98:
@mautic/core-team can you help to review the content here? TIA! ✨

escopecz:
Once https://github.com/mautic/mautic/pull/14019 is merged then this will need another overhaul. Perhaps we could wait for that?

docs/campaigns/campaign_builder.rst
RCheesley: Yes, I think that's fine, using a link macro

adiati98:
@RCheesley the build failed here because of timeout from this newly added URL: https://kb.mautic.org/article/how-to-integrate-mautic-with-salesforce.html.

Should we include this in the `accept.txt`?

RCheesley:
I reran and it works so all good!

PR #174, opened 2023-06-19
Symfony mailer docs

escopecz:
Docs for https://github.com/mautic/mautic/pull/11613

More details for upgrading from M4 and developer changes are documented in `UPGRADE-5.0.md`

https://github.com/mautic/mautic/pull/11613/files


escopecz:
@RCheesley @mabumusa1 @fedys please review this doc and let me know if I forgot something

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
 Email delivery
 ##############
 
-Mautic delivers Emails using the method defined by the system administrator. If you are the system administrator for your Company, then you need to add the Email protocol for your Mautic instance to use. Mautic integrates with any Email service provider which offers SMTP mail servers as well as several distinct services such as:
+Mautic delivers Emails using the method defined by the system administrator. If you are the system administrator for your Company, then you need to add the Email protocol for your Mautic instance to use. Mautic integrates with any Email service provider which offers SMTP mail servers. As Mautic uses the :xref:`Symfony Mailer` library since v5, it supports its core plugins. Transports for other email services might be found on Github or Packagist.
```
RCheesley, line 228: Vale fix and making it clearer about the core plugins.

```suggestion
As Mautic uses the :xref:`Symfony Mailer` library since v5, it supports all Symfony Mailer core plugins out of the box. Transports for other Email services might be found on GitHub or Packagist.
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
 
-- :xref:`Mandrill`
-- :xref:`GMail`
-- :xref:`Sendgrid`
-- :xref:`Mailjet`
-- :xref:`Postmark`
-- :xref:`Sendmail`
-- :xref:`Amazon SES`.
+SMTP Transport
```
RCheesley, line 230: Vale fix

```suggestion
SMTP transport
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+**************
+
+The SMTP transport is the default transport for Mautic. It's configured in the Mautic configuration under the Email Settings tab. The configuration is the same as in the :xref:`Symfony Mailer` documentation.
```
RCheesley, line 233: Making it a bit clearer for the end-user.

```suggestion
The SMTP transport is the default transport used for sending Emails with Mautic. It's configured in the Mautic configuration under the Email Settings tab. The configuration is the same as in the :xref:`Symfony Mailer` documentation.
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+This is the example DSN mentioned in the :xref:`Symfony Mailer` documentation for SMTP:
```
RCheesley, line 235: First time we use a new abbreviation, we should use the full-text.

```suggestion
This is the example Delivery Status Notification (DSN) configuration mentioned in the :xref:`Symfony Mailer` documentation for SMTP:
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+.. code-block:: shell
+    
+    smtp://user:pass@smtp.example.com:port
+
+This is how this DSN converted to Mautic's email configuration:
```
RCheesley, line 241: I think this is what you mean?

```suggestion
Mautic creates this automatically from the values entered in the Email configuration:
```

@escopecz please add a table which explains what each of the values actually means. Should just need three columns I guess or maybe two - value name, what it means, any notes / links / references.  Docs on how to do the markdown are here: https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html#tables. Thanks!

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+.. image:: images/emails/smtp-dsn.png
+    :width: 400
+    :alt: SMTP API DSN example
+
+.. note::
+  Use the Mautic's global configuration to paste in the DSN information. Especially the API keys and passwords. The values must be URL-encoded and the configuration form will do that for you. If you are pasting DSN directly into the config/local.php file, you must URL-encode the values yourself.
```
RCheesley, line 248: Vale fix.

```suggestion
  Use the Mautic's global configuration to paste in the DSN information, especially the API keys and passwords. The values must be URL-encoded, and the configuration form does that for you. If you are pasting DSN directly into the config/local.php file, you must URL-encode the values yourself.
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+Example API Transport Installation
```
RCheesley, line 250:
```suggestion

.. vale off

Example API transport installation

.. vale on

```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+**********************************
+
+Let's say you want to use :xref:`Sendgrid` API instead of SMTP to send Emails. You can install the official Symfony Sendgrid Transport by running the following command that is mentioned along others in the :xref:`Symfony Mailer` documentation.
```
RCheesley, line 253: Vale fix

```suggestion
If you want to use :xref:`Sendgrid` API instead of SMTP to send Emails, for example, you can install the official Symfony Sendgrid Transport by running the following command that is mentioned along others in the :xref:`Symfony Mailer` documentation.
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+.. code-block:: shell
+    
+    composer require symfony/sendgrid-mailer
```
RCheesley, line 257: What about people who are not using Composer? How do they use it?

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+After that, you can configure the transport in the Mautic configuration. The example DSN is again mentioned in the :xref:`Symfony Mailer` documentation along with other transports. In case of Sendgrid API, the DSN would look like this:
```
RCheesley, line 259:
```suggestion
After that, you can configure the transport in the Mautic configuration. The example DSN is again mentioned in the :xref:`Symfony Mailer` documentation along with other transports. In the example of using the Sendgrid API, the DSN looks like this:
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+.. code-block:: shell
+    
+    sendgrid+api://KEY@default
+
+This is how this DSN converted to Mautic's email configuration:
```
RCheesley, line 265:
```suggestion
This is how it would be set up in Mautic's Email configuration:
```

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+
+  .. image:: images/emails/sendgrid-api-dsn.png
+    :width: 400
+    :alt: Sendgrid API DSN example
+
+Once you replace the Sendgrid API KEY, you can add it to the DSN and save the configuration. Mautic will now use the Sendgrid API to send Emails. 
+
+.. note::
+  It is nice perk that Mautic can use any Symfony Mailer transport. However, be aware that such transport do not support batch sending even via API. They only send one email per request opposed to a thousand emails per request as some special Mautic tansports can. It also does not support transport callback handling. If you plan to send bigger ammounts of emails please consider using email transports built specifically for such use. These plugins are available in :doc:`Mautic Marketplace </marketplace/marketplace.rst>`.
```
RCheesley, line 274:
```suggestion
  It's nice perk that Mautic can use any transport provided by Symfony Mailer. However, be aware that such transports (from Symfony) don't support batch sending, even via API. They only send one email per request, as opposed to a thousand emails per request as is the case with some Mautic transports, which can make them slow at scale. They also don't support transport callback handling used for bounce management and reply detection. If you plan to send larger volumes of Emails or need to use features which require callback handling, please consider using Email transports built specifically for such use. These plugins are available in the :doc:`Mautic Marketplace </marketplace/marketplace.rst>`.
```

docs/channels/emails.rst
```diff
@@ -249,7 +287,7 @@ Mautic works most effectively with high send volumes if you use the queued deliv
 
 .. code-block:: shell
     
-    php /path/to/mautic/bin/console mautic:email:process
+    php /path/to/mautic/bin/console messenger:consume email_transport
```
RCheesley, line 290: Please update also the cron jobs file, and the command line interfaces file. Both will need to have this command included and explained.

RCheesley (changes requested):
Thanks for making the PR @escopecz and for the detailed information. Some Vale fixes and suggestions to improve readability for the user.

RCheesley (changes requested):
Thanks for making the PR @escopecz and for the detailed information. Some Vale fixes and suggestions to improve readability for the user.

docs/channels/emails.rst
```diff
@@ -225,15 +225,53 @@ Marketers can place the signature into an Email using the ``{signature}`` token.
...
+  .. image:: images/emails/sendgrid-api-dsn.png
+    :width: 400
+    :alt: Sendgrid API DSN example
+
+Once you replace the Sendgrid API KEY, you can add it to the DSN and save the configuration. Mautic will now use the Sendgrid API to send Emails. 
+
+.. note::
+  It is nice perk that Mautic can use any Symfony Mailer transport. However, be aware that such transport do not support batch sending even via API. They only send one email per request opposed to a thousand emails per request as some special Mautic tansports can. It also does not support transport callback handling. If you plan to send bigger ammounts of emails please consider using email transports built specifically for such use. These plugins are available in :doc:`Mautic Marketplace </marketplace/marketplace.rst>`.
 
 The system can either send Emails immediately or queue them for processing in batches by a :doc:`cron job </configuration/cron_jobs>`.
```
mabumusa1, line 276: I think we should kill the immediate mode from the docs and ask users to use the queue mode all the time

mabumusa1 (changes requested):
I think queue method should be the ONLY method to send out emails either it is batch or single emails. there is no point of using in memory sends

docs/channels/emails.rst
escopecz: I don't think it makes sense to use those without Composer. Non-Composer Mautic installations will have to go for a Mautic plugin that does not have any other dependencies if SMTP is not enough.

docs/channels/emails.rst
escopecz: Will do

docs/channels/emails.rst
RCheesley: I think that we need to make this clear up front then, that you need to be using composer? Or will be be supplying such plugins?

docs/channels/emails.rst
escopecz: The immediate mode is using `sync://` messenger option, right? Or are you talking about something else?

mabumusa1: Yes, that is right. I suggest we do not use sync at all

escopecz: I thin it's useful when someone is just testing Mautic out or have Mautic with a few dozens of contacts. @RCheesley WDYT?

docs/channels/emails.rst
escopecz: I added a warning in https://github.com/mautic/user-documentation/pull/174/commits/e1850808dc6e8362c6338ec051280e2dd80c8e50

docs/channels/emails.rst
RCheesley: I do think that immediate mode has a place for low-volume sending or when the mails really do need to go out immediately.

I'd only be in favour of killing it if we're absolutely sure that every single user of Mautic on any hosting environment is going to be capable and willing to use the queueing system, and we can document how to make that effectively send the emails as if they were immediate.

docs/channels/emails.rst
escopecz: The table was added in https://github.com/mautic/user-documentation/pull/174/commits/7411e6172d793b1b2b1345ddfaf6192b8248b2e5

escopecz:
@RCheesley do I have to rebased this on the 5.x branch?

RCheesley:
> @RCheesley do I have to rebased this on the 5.x branch?

For things which are 5.x specific, yes please. I need to do a back-merge as haven't done it since November, I can look at doing that later in the week.

RCheesley:
Safe to ignore the build error as we've excluded the Twilio domain. Thanks @escopecz 🚀

PR #177, opened 2023-06-27
Move Email config / DSN settings to Configuration section

RCheesley:
This PR moves the configuration for the email settings / DSN over to the email config part, where it should really live.

I also fixed a ton of incorrect apostrophes which is common when copy/pasting from another source.

<a href="https://gitpod.io/#https://github.com/mautic/user-documentation/pull/177"></a>


docs/configuration/settings.rst
```diff
@@ -161,6 +161,114 @@ Campaign settings
 Email settings
 **************
 
+Email transport settings
```
RCheesley, line 164: @escopecz I figured that email transport settings might be a more accurate heading, WDYT?

escopecz (dismissed):
Makes sense. Thanks!

escopecz:
@RCheesley can we merge this? @fedys could use this as a base for the queue setting docs.

escopecz:
Another question, shouldn't this be against the 5.x branch?

RCheesley:
Ah yeah thanks for that @escopecz ... I smooshed up the rebase but think it's fixed now.

The outstanding vale issues are warnings/notices which are not relevant in my opinion so we should be OK to go with this now.

PR #182, opened 2023-07-11
add step for MyST Parser

diaboloshogunate:
In addition to the install steps I had to install MyST Parser before I could run `make html` without failing.

<a href="https://gitpod.io/#https://github.com/mautic/user-documentation/pull/182"></a>


RCheesley (approved):
Thanks for adding the info @diaboloshogunate 🚀

PR #194, opened 2023-08-04
email_transport queue is now called just email

escopecz:
When calling `bin/console messenger:consume email_transport` I'm getting this error:

```
The receiver "email_transport" does not exist. Valid receivers are: email, failed.
```

So it should be just `email`


RCheesley (approved):
Thanks for the quick fix! GTG 🚀

PR #178, opened 2023-06-29
Messenger configuration docs

fedys:
Docs for https://github.com/mautic/mautic/pull/12509

<a href="https://gitpod.io/#https://github.com/mautic/user-documentation/pull/178"></a>

PR #195, opened 2023-08-10
draft messenger documentation

galvani:
Docs for https://github.com/mautic/mautic/pull/12314

<a href="https://gitpod.io/#https://github.com/mautic/user-documentation/pull/195"></a>


docs/messenger/messenger.rst
```diff
@@ -0,0 +1,122 @@
...
+The default configuration is as follows
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: php
+
+    'routing' => [
+        \Symfony\Component\Mailer\Messenger\SendEmailMessage::class => MauticMessengerTransports::EMAIL,
+        \Mautic\MessengerBundle\Message\PageHitNotification::class  => MauticMessengerTransports::HIT,
+        \Mautic\MessengerBundle\Message\EmailHitNotification::class => MauticMessengerTransports::HIT,
+    ],
```
escopecz, line 61: What is this information for? Will users need to configure it themselves or change it?

docs/messenger/messenger.rst
```diff
@@ -0,0 +1,122 @@
...
+-------
+
+Every message needs to have a transport associated with it. The synchronous transport is a valid selection. By default, PageHit and EmailHit messages are routed to the **hit** transport, which is configured as synchronous.
+
+.. code-block:: php
+
+    MauticMessengerTransports::HIT    => [
+        'dsn'        => 'sync://',
+        'serializer' => 'messenger.transport.jms_serializer',   // Smaller payload
+    ],
```
escopecz, line 48: Where should users put this? Or isn't this the default behavior?

docs/messenger/messenger.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Replace ``your_transport`` with the name of your transport.
+
+To consume messages from all transports, you would use:
+
+.. code-block:: bash
+
+    php bin/console messenger:consume
+
+The command will keep running and consume messages as they come until it's manually stopped.
```
escopecz, line 85: I'm getting this message when I try this:
```

www-data@28de97d2dce5:~/html$ bin/console messenger:consume

                                                                                                                        
 Which transports/receivers do you want to consume?                                                                     
                                                                                                                        

Choose which receivers you want to consume messages from in order of priority.
Hint: to consume from multiple, use a list of their names, e.g. email, failed

 Select receivers to consume: [email]:
  [0] email
  [1] failed
```

escopecz (changes requested):
Thanks Jan! It loos good! I have a few questions. Also, please rebase this against the `5.x` branch.

docs/messenger/messenger.rst
galvani: seems like i have to verify everything ai says :-( sorry

docs/messenger/messenger.rst
galvani: it is to show how routing is done so it is easier to understand why we configure just the one transport a little lower. so they understand the routing
message => route => transport => handler

galvani:
I have fixed the issues and updated.

docs/messenger/messenger.rst
galvani: I have described it better now.

escopecz (commented):
It seems there are unrelated comments in this PR now. But looking at them I think the new doc will have to be registered in docs/index.rst too. I'm also still new to this new doc. @RCheesley might tell for sure.

RCheesley:
This PR still needs rebasing to 5.x please? Thanks!

docs/messenger/messenger.rst
```diff
@@ -0,0 +1,131 @@
...
+.. _highlights:
+
+**Load distribution/control**
+    It enables you to balance and control the server load effectively.
+
+**Improved client response**
+    By processing requests in the background, it ensures that the client receives a response before the completion of the request processing.
+
+**Scalability**
+    It provides the flexibility to scale your Mautic instance by adding more workers.
```
RCheesley, line 15: Will marketers know what workers are? I don't know if we have explained that concept at all before.

docs/messenger/messenger.rst
```diff
@@ -0,0 +1,131 @@
...
+=========
+
+Messenger is a feature in Mautic 5.x that offers several significant benefits:
+
+.. _highlights:
+
+**Load distribution/control**
+    It enables you to balance and control the server load effectively.
+
+**Improved client response**
```
RCheesley, line 11: What are we referring to here when we are talking about clients? Will marketers understand that?

RCheesley (changes requested):
Thanks for making a start on the docs @galvani !

Some general feedback:

1. Please move messenger/messenger to configuration/messenger - that is where a user would go to find information about configuring the email / messaging systems rather than a totally separate section.

2. Please update index.rst to include it in the section starting at line 40 as follows:

```
.. toctree::
   :caption: Configuration
   :maxdepth: 2
   :hidden:

   configuration/maxmind_license
   configuration/command_line_interface
   configuration/cron_jobs
   configuration/variables
   configuration/settings
   configuration/messenger
```

This adds the file to the navigation structure.

3. Please install the Vale linter - it is available for PHPStorm and VSCode (and most other IDEs). This will help to pick up grammar issues as you type and help guide you to fix them. Ask for help if you need it with things like passive->active speech as that is hard to do even for a native speaker (ChatGPT can help sometimes!)

4. Any settings that relate to Messenger from the UI (if any - I wasn't clear on whether any of this is visible in the user interface) should be added to Configuration > Settings - this is where we run through every setting, option and switch in the configuration screens. So explain what they see in the user interface in that file. Link to messenger.rst for more details on what it actually means.

5. We have to consider that this is the user documentation, for predominantly _marketers_. I think we need a basic overview of what people _need_ to know to use Mautic out of the box - what is configured by default and where, what are the default settings in the config (always useful if you screw something up and need to revert back without the undo functionality!) etc.  Then we should have an advanced section which dives into more detail.

 I've asked some users to take a look and give some feedback - we will help to work collaboratively on a structure for the information that will work for all users!

escopecz:
@fedys made an update to @galvani 's PR in https://github.com/mautic/mautic/pull/12648 which simplifies the configuration and still enables the option to overwrite it in config-local.php with more advanced configuration if needed. If that PR is accepted, I'll update these docs accordingly.

RCheesley:
Closing in favour of #178

PR #206, opened 2023-10-27
Fix vale action

RCheesley:
This fixes a bug reported with Vale/Reviewdog which required updating: https://github.com/errata-ai/vale/issues/721.

Actions are now working and reporting issues as found, so should be good to go! 🚀


escopecz (approved):
Looks good to me. Thanks Ruth! 👍

PR #219, opened 2023-12-24
Fix build fail

RCheesley:
Turns out the readthedocs.yml file had a space after the filename which is causing build failures.

Also removed the prebuilds section of the config as that is deprecated and managed within the UI of Gitpod now.


mollux (approved):
nice catch :) changes look good

RCheesley:
Entirely thanks to RTD support, I was completely perplexed as everything seemed to be as it should be! 🤦

PR #235, opened 2024-01-10
Add troubleshooting section

### LINKED ISSUE #225: Remove reference to troubleshooting section
We removed the general troubleshooting section, but this page still refers to it:

https://docs.mautic.org/en/5.x/getting_started/how_to_update_mautic.html#the-update-wasn-t-successful

This should be updated (and maybe we need to point them to the forum? Or maybe we need a troubleshooting section? Open to thoughts!)

RCheesley:
This PR fixes #225 by adding back the missing Troubleshooting section.

Future PRs might move the troubleshooting pages from other sections into this, but for now this will help with the immediate issue post-new-version.


docs/troubleshooting/troubleshooting.rst
```diff
@@ -0,0 +1,141 @@
...
+4. Try to update the files manually
+===================================
+
+This step requires some manual intervention - there is no command for this part.
+
+To update the files manually, you have to:
+
+   1. Back up (download) all Mautic files from your server to your local computer, using File Transfer Protocol (FTP) or the ``scp`` command, which is much faster.
+   2. Delete all Mautic files and folders on your server. Use FTP or the rm command - use the latter with extreme caution.
+   3. Download the latest Mautic package from :xref:`Mautic Download`.
```
escopecz, line 89: This could be replaced with a link to the Composer installation

escopecz (approved):
It looks great. One suggestion but I think it can be merged as is and then update it with the composer docs updates that will come soon.

docs/troubleshooting/troubleshooting.rst
RCheesley: OK let's update this once our Composer docs are more streamlined and user friendly!

PR #237, opened 2024-01-10
Add Troubleshooting to menu

RCheesley:
Seems there is some issue with the way I added it to the index. Need to figure out what the problem is as it's not showing on the menu.


RCheesley:
This is good to go - I accidentally duplicated one of the settings which broke the world!

screenshot-mautic-userdocumentatio-3482cvuced0 ws-eu107 gitpod io-2024 01 10-16_55_49

PR #238, opened 2024-01-15
Bump RTD Sphinx Search

RCheesley:
Per the advisory: https://github.com/readthedocs/readthedocs-sphinx-search/security/advisories/GHSA-xgfm-fjx6-62mj

We should update the dependency.


RCheesley:
I've tested this locally (as you can't test it on Gitpod) and search is working as expected.

fakela (approved):

fakela:
> I've tested this locally (as you can't test it on Gitpod) and search is working as expected.

That's great 😃

PR #252, opened 2024-03-05
---

andersonjeccel:


LordRembo (changes requested):
You made changes to the docs, that cover different scopes. Eg. replacing 'cog icon' with 'list items icon'. But also stuff to do with the 'published' state labels.
That is a dangerous thing. Scopes should be separated
 In this particular case, we decided to revert to the 'cog' icon, so these references need to be fixed

RCheesley:
Closing this PR re the comments above.

PR #255, opened 2024-03-15
Move tracking script from within Cron Jobs

### LINKED ISSUE #224: Move tracking script from within Cron Jobs
Not sure how, but the tracking script instructions seem to have been added on the cron jobs page which seems a bit counterintuitive.

https://docs.mautic/org/en/5.x/configuration/cron_jobs.html#tracking-script

I think this should be under the installing Mautic section, perhaps?

NOTE: Update redirect when this is done.

nox1134:
Fixes #224 by @RCheesley 
It appears that the tracking script instructions have been added to the cron jobs page.  So, moving tracking script from within the cron jobs page to it's own page.
image
I'm not sure about what you mean by updating the redirect in the issue? (If it means updating in the readthedocs, then I think you need to do this through readthedocs dashboard)


RCheesley:
Thanks for the PR! 

The note about the redirect was a 'note to self/team' that when this is done we'd need to redirect the old URL to the new location. You don't need to add those to the document, but instead please pop it in a comment here so we make sure to add the redirect on Read the Docs once the PR is merged!

If you can just remove that section we can then get this merged.

RCheesley (approved):
Good to go, thanks for the PR!

PR #258, opened 2024-03-20
[UX] Change how the ways to implement a form are called

andersonjeccel:
This PR changes docs for https://github.com/mautic/mautic/pull/13476


RCheesley (approved):
Thanks for the update @andersonjeccel - I have added a new milestone to the docs so we know which PRs need to be merged when 5.1 is released.

cc @fakela @favour-chibueze for info.

PR #272, opened 2024-04-16
Change "System Report?" to "Visible for other users"

andersonjeccel:
Related to https://github.com/mautic/mautic/pull/13496


docs/reports/reports.rst
```diff
@@ -88,9 +88,9 @@ The parent data source provides a high-level summary of the data while the child
   
 * **Published** - Select **Yes** to ensure that the Report is active, ready to send when scheduled in the Report options.
 
-* **System Report** - Select **Yes** so that all Users can access the Report. If set to No, only the owner of the Report and Users with permission to view others' Reports can see the Report.
+* **Visible for all logged-in users** - Select **Yes** so that all Users can access the Report. If set to No, only the owner of the Report and Users with permission to view others' Reports can see the Report.
```
RCheesley:
```suggestion
* **Visible for all logged-in Users** - Select **Yes** so that all Users can access the Report. If set to No, only the owner of the Report and Users with permission to view others' Reports can see the Report.
```

docs/reports/reports.rst
```diff
@@ -88,9 +88,9 @@ The parent data source provides a high-level summary of the data while the child
...
 
-* **Owner** - Select the owner of the Report to determine who can see the Report if the System Report setting is No.
+* **Owner** - Select the owner of the Report to determine who can see the Report if the Visible for all logged-in users setting is No.
```
RCheesley:
```suggestion
* **Owner** - Select the owner of the Report to determine who can see the Report if the 'Visible for all logged-in users' setting is No.
```

RCheesley (changes requested):
Small Vale fix and a suggestion on putting the setting reference in single quotes. Thanks @andersonjeccel 🚀

docs/reports/reports.rst
andersonjeccel: Is this a standard?
Or we are trying to make it standard?

RCheesley: In the docs we always capitalise features of the product - we apply the Google Technical Writing Style Guide: https://developers.google.com/style - although it seems there was a recent update which now says _not_ to capitalise features where it used to say that you should!

For the time being we stick with what we currently use.

docs/reports/reports.rst
```diff
@@ -88,7 +88,7 @@ The parent data source provides a high-level summary of the data while the child
   
 * **Published** - Select **Yes** to ensure that the Report is active, ready to send when scheduled in the Report options.
 
-* **Visible for all logged-in users** - Select **Yes** so that all Users can access the Report. If set to No, only the owner of the Report and Users with permission to view others' Reports can see the Report.
+* **Visible for all logged-in Users** - Select **Yes** so that all Users can access the Report. If set to No, only the owner of the Report and Users with permission to view others' Reports can see the Report.
 
 * **Owner** - Select the owner of the Report to determine who can see the Report if the 'Visible for all logged-in users' setting is No.
```
RCheesley:
```suggestion
* **Owner** - Select the owner of the Report to determine who can see the Report if the 'Visible for all logged-in Users' setting is No.
```
Looks like this one needs changing too according to Vale!

RCheesley (changes requested):

RCheesley (approved):

PR #279, opened 2024-04-16
add audit log in the list

fakela:


docs/reports/reports.rst
```diff
@@ -39,17 +39,21 @@ The **Details** tab on a Report contains the same options across all Reports and
    
    * Message Queue
 
-4. :doc:`Emails</channels/emails>`
+4. Audit Log
```
RCheesley, line 42: Should this have page of its own given the others do? Guessing we might have to think where it would sit - maybe under Configuration?

RCheesley (commented):
I think that there should probably be a page which explains what the audit log is and why/how you would use it, perhaps?

RCheesley:
@fakela this one is pending feedback - I think we should have a page which explains the audit log to accompany this PR.

adiati98:
@fakela can you please response to the pending feedback here? TIA! ✨

PR #289, opened 2024-04-16
Add new command reference #10775

RCheesley:
Adds details for new command to analyze custom fields for the purpose of optimization.

Ref https://github.com/mautic/mautic/pull/10775.


escopecz (dismissed):
Thank you, Ruth! This is correct 👍 

It may be worth saying that this way users can optimize the VARCHAR columns so they could create more custom fields if they hit the hard limit on the leads table and cannot create more.

RCheesley:
> It may be worth saying that this way users can optimize the VARCHAR columns so they could create more custom fields if they hit the hard limit on the leads table and cannot create more.

Please check the latest commit - I added some clarifying text on this. Thanks for the review!

escopecz (dismissed):
Awesome, thanks! 👍

PR #292, opened 2024-04-17
[UX] Reorder administration panel

andersonjeccel:
For https://github.com/mautic/mautic/pull/13596


docs/plugins/images/plugins.png
RCheesley, line 1: Is there a reason to change the colour used? As we tend to stick with red.

andersonjeccel: Hm, no reason (it was the default of the Photos app)
Would it be a problem?

RCheesley, line 1: Could we have this one with a red line rather than blue please @andersonjeccel , to keep in context with the ret of the docs?

RCheesley (changes requested):
Just need to update the image so it's using a red underline instead of blue, please!

docs/plugins/images/plugins.png
andersonjeccel: @RCheesley Fixed

RCheesley (approved):

PR #296, opened 2024-04-24
#13368 update form labels

andersonjeccel:
update for https://github.com/mautic/mautic/pull/13368


docs/components/forms.rst
```diff
@@ -105,7 +105,7 @@ To add a new field to your Form:
 
    - **Hidden** - This field won't be visible on the Form, but include default values, saved along with the Form submission, for reporting or internal tagging purposes.
 
-   - **List - Country** - This populates Mautic's default, non-editable country list. To use a custom list you should make use of the Select field type and manually enter the countries you would like to include.
+   - **Select: Country** - This populates Mautic's default, non-editable country list. To use a custom list you should make use of the Select field type and manually enter the countries you would like to include.
```
RCheesley, line 108:
```suggestion
.. vale off

   - **Select: Country** - This populates Mautic's default, non-editable country list. To use a custom list you should make use of the Select field type and manually enter the countries you would like to include.
   
  .. vale on
```

Vale is moaning about this so we can turn it off temporarily with this syntax. Just ensure that it has a line break above and below each statement (couldn't really see if that was the case with the suggestion). we can do that with each of the bullets below.

RCheesley (changes requested):
@andersonjeccel FYI when linking to a PR outside of the repo, please use the full link not just the PR number, as it does not have the context and therefore doesn't link them.

I've suggested switching Vale off for these because we want to break our own rules based on how we've got it written in Mautic, so we should tell Vale to ignore these lines.

docs/components/forms.rst
andersonjeccel: Hm, I didn't know we had a test for it
Should I try to fit the guidelines or just disable?

RCheesley: We can't, because the form is actually using the syntax stated in the dropdown option. Hence suggesting just turning Vale off for this section and then back on again.

PR #297, opened 2024-04-26
Add detail to Composer docs

RCheesley:
I think a lot of folks are having issues with Composer based installations because we don't make it clear about the implications of Mautic being served from /docroot by default.

This means that you must update your root directory on your hosting, and I think a lot of folks (myself included the first few times) stumble over this point. It results in a bunch of errors about autoloader not being found etc.

It's a small update but hopefully it will be the missing piece of the puzzle for a few people.

Also a small update to a later command which was still using ^4 instead of ^5


favour-chibueze (approved):

RCheesley:
Thanks for the review @favour-chibueze - let's get this merged, I've asked multiple times for a check from composer peeps but we can always revert the PR if there's any problems.

PR #303, opened 2024-05-07
Hide/Unhide themes documented

escopecz:
Change that documents https://github.com/mautic/mautic/pull/13723


escopecz:
I'm not sure how to deal with the last 2 Vale issues. Any suggestions?

docs/themes/manage_themes.rst
```diff
@@ -104,6 +104,31 @@ To preview a Theme:
...
+
+2. Click the **Settings** icon.
+
+3. Click **Themes**.
+
+4. In the Themes section, locate the Theme that you want to hide.
+
+5. Select the drop-down before the Theme, and click **Hide**.
+
+6. Mautic moves the Theme to the bottom of the table and show it in grey. This means the Theme is hidden in the builders.
```
RCheesley:
```suggestion
6. Mautic moves the Theme to the bottom of the table and show it in grey, which hides it within builders.
```

Vale fix

docs/themes/manage_themes.rst
```diff
@@ -104,6 +104,31 @@ To preview a Theme:
...
+
+To revert this change open the context menu of the hidden Theme and click **Unhide**.
```
RCheesley, line 129: This should really be 'reveal' or something like that - unhide isn't really a great term to use. Is this something we have in Core already @escopecz ? If so we can wrap this line in a vale off/on statement.

RCheesley (changes requested):
One fix, and one query about the terminology used.

docs/themes/manage_themes.rst
escopecz: Yes, it's in the core already

docs/themes/manage_themes.rst
```diff
@@ -104,6 +104,31 @@ To preview a Theme:
...
+2. Click the **Settings** icon.
+
+3. Click **Themes**.
+
+4. In the Themes section, locate the Theme that you want to hide.
+
+5. Select the drop-down before the Theme, and click **Hide**.
+
+6. Mautic moves the Theme to the bottom of the table and show it in grey, which hides it within builders.
+
```
RCheesley, line 128:
```suggestion
.. vale off

```

docs/themes/manage_themes.rst
```diff
@@ -104,6 +104,31 @@ To preview a Theme:
...
+To revert this change open the context menu of the hidden Theme and click **Unhide**.
+
```
RCheesley, line 131:
```suggestion
.. vale on

```

RCheesley (changes requested):
Let's wrap that bit in a Vale Off so it doesn't flag the error. I'll make a PR to suggest a change to the wording.

The vale on/off needs a line break above and below, it wasn't totally clear from the GitHub UI whether that works with my suggestions so please just make sure the spacing comes out alright! :)

RCheesley:
https://github.com/mautic/mautic/pull/13835 PR to improve the grammar here :)

docs/themes/manage_themes.rst
RCheesley: Yep, made the PR to address that :)

PR #307, opened 2024-06-13
Fix failing build

### LINKED ISSUE #305: Fix broken table on command line page
https://docs.mautic.org/en/5.x/configuration/command_line_interface.html this page should show a list of the commands with their aliases but it seems to be empty. This seems to be because it's broken after the scheduled export command: https://github.com/mautic/user-documentation/blob/5.x/docs/configuration/command_line_interface.rst.

Should be a quick fix!

RCheesley:
This PR:

- Fixes #305 
- Fixes other links in content which were broken and causing a failing build
- Fixes a duplicated heading
- Fixes a warning about bold text not being closed


nick-vanpraet (approved):
Looks good

PR #308, opened 2024-06-13
Update composer update documentation

nick-vanpraet:
Also updated vale documentation link.

I believe I did not tell a lie, but best to have a few eyes on it. I left in all the git stuff, even though nothing mentions you should use git or what it even is but I felt that was out of scope for now.


docs/getting_started/how_to_update_mautic.rst
```diff
@@ -33,15 +33,23 @@ If you customize any of the 'scaffolding' files - commonly ``.htaccess`` - you m
...
 Follow the steps below to update your core files.
 
-1. Run ``composer update mautic/core-lib --with-dependencies`` to update Mautic core and its dependencies.
+1. Backup your ``composer.lock`` and ``composer.json`` file. If something does not work as expected during the ``composer update`` command, restore them and run ``composer install`` to return your codebase to the state it was in before attempting to update.
 
-2. Run ``git diff`` to determine if any of the scaffolding files have changed. Review the files for any changes and restore any customizations to ``.htaccess`` or others.
+2. Edit the ``composer.json`` file, and change all previous versions with the version you wish to update to for all Mautic packages.
 
-3. Commit everything all together in a single commit, so the ``docroot`` remains in sync with the core when checking out branches or running ``git bisect``.
+    * If you are running ``5.0.4`` and want to update to ``5.1.0``, replace ``5.0.4`` with ``5.1.0`` for all packages that start with ``mautic/`` and are set to ``5.0.4``.
```
RCheesley:
```suggestion
    * If you are running ``5.0.4`` and want to update to ``5.1.0``, replace ``5.0.4`` with ``5.1.0`` for all packages that start with ``mautic/`` and currently use ``5.0.4``.
```

Vale fix

docs/getting_started/how_to_update_mautic.rst
```diff
@@ -33,15 +33,23 @@ If you customize any of the 'scaffolding' files - commonly ``.htaccess`` - you m
...
+    * You may also need to increase the versions of any other packages you have either manually added or added through :ref:`Mautic Marketplace`.
+    * If you have not added any extra packages, you can also replace the entire ``composer.json`` file with the newer version from the `repository <https://github.com/mautic/recommended-project>`_ that matches your desired target version.
```
RCheesley:
```suggestion
    * If you haven't added any extra packages, you can also replace the entire ``composer.json`` file with the newer version from the `repository <https://github.com/mautic/recommended-project>`_ that matches your desired target version.
```

Vale fix

RCheesley (changes requested):
Some vale fixes based on the test failures.

Does this also hold true for people using Composer with Recommended Project? Do we also need to warn about for example htaccess which might have custom modifications?

nick-vanpraet:
> Does this also hold true for people using Composer with Recommended Project? Do we also need to warn about for example htaccess which might have custom modifications?

This is specifically for people using Recommended Project. Which is the only mention I see in the setup doc pages, so I figure that is the only one we officially support. Anyone who adds packages themselves to their own composer they setup on their own should know enough to manage regardless.

The warnings about git diff and .htaccess are all still present, I just added 2 extra bullet points before the existing ones and rewrote the one that said to run `composer update mautic/core-lib`.

RCheesley (approved):
I do think we should clarify the bit about using git diff and when it would be used, but that can be done in another PR.

PR #311, opened 2024-06-14
Update gmail.rst

DJUserkent:
the plugin no longer works - I've added text to say that.


RCheesley:
Thanks for the PR @DJUserkent - do you think it might be better just to remove this page all together?

DJUserkent:
It's your call Ruth.

My hunch would be to leave it, on the principle that its helpful to
leave an authoritative 'archeological' trail.

Because folks in the future can still find online content that mentions
this -  in the forum and github and maybe other people's websites. So
they will wonder ' what is the status of this' and ask again in Forums
at regular intervals. And people will speculate.

Whereas an authoritative mautic page: allows people to find the true
situation more easily = less unhelpful questions in the forums.

*But a good idea*

would be to change the page title  - so that the title already states
the main fact, without needing to open the page and see the top line.

Again - that saves the reader some time -and means the 'Table of
Contents' type lists anywhere, will be more helpful.

RCheesley:
I'll leave this up to the @mautic/education-team-leaders to decide, I feel like if it's something which is broken then it shouldn't be in our docs until it's in a usable state. As far as I know this hasn't worked for at least 5+ years, we've tried reaching out to the developer who built it to ask if they could update it to no avail. The GitHub repo is still there for people to find, however.

DJUserkent:
Actually Ruth

thinking about Mautic documentation - maybe it is my brain, but I have
googled the Mautic documentation and forums alot - but am struggling to
get the Email working.  And this month there has been an announcement on
V5 that the email handling is changing, in favour of  Symphony handling
it. ie dropping the ability to API to (eg) Sendgrid.

Perhaps there is one page that sets it all out clearly - but if so, I've
not found it.

The challenge we have is around fails / bounces

*    on 4.4.

*    we used the API to send via Sendgrid

*    messages get sent fine - but where the receiving MTA rejects the
email before receiving it:   SendGrid over the API should be telling
mautic: but doesn't seem to?    There is no 'bounce email' in this case,
as the receiving MTA never accepted the email at all in the SMTP
exchange. (bounces only occur if the MTA received it, but then minutes
later sends an email bounce ' sorry, actually, we don't have that mailbox')

fakela:
> Thanks for the PR @DJUserkent - do you think it might be better just to remove this page all together?

We can remove this page and probably have a page for list of plugins that no longer works in case anyone wants to verify from the docs

favour-chibueze (dismissed):

RCheesley:
I think we should remove the page entirely. The docs will still be there in the 4.x version for reference in case someone needs to get them.

PR #312, opened 2024-06-17
Update Vale rules to prompt re. published / unpublished no longer being used

RCheesley:
This PR updates the Vale featurelist which prompts people to use the correct terminology.

It encourages them to use Activate / Activated / Turn On or Deactivate / Inactive / Turn Off.


RCheesley:
@andersonjeccel here's a PR to update the prompts to take into account the changes to published/unpublished, which should help us update any places we've missed and prompt people going forward to use the correct terms.

andersonjeccel:
Really good one, thanks!

PR #337, opened 2024-10-22
Fix linter

RCheesley:
This PR installs Python + dependencies prior to running the Vale linter, which should fix the broken GitHub Actions.


RCheesley:
Looks like the build failure is a timeout on Webmecaniks' site so this should be good to get merged.

PR #352, opened 2025-03-07
Documenting the custom field delete command

escopecz:
Docs for https://github.com/mautic/mautic/pull/14679

### DIFF
--- a/docs/contacts/custom_fields.rst
+++ b/docs/contacts/custom_fields.rst
@@ -102,7 +102,7 @@ Since :xref:`Mautic 3` there is an option you can set in your ``app/config/local
 
 If you configure this option, the new Custom Field becomes visible in the list of Custom Fields. The Custom Field remains unpublished until you run the command ``bin/console mautic:custom-field:create-column``. This command creates the actual column in the table and publishes the field metadata.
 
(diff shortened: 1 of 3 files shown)


docs/contacts/custom_fields.rst
```diff
@@ -102,6 +102,8 @@ Since :xref:`Mautic 3` there is an option you can set in your ``app/config/local
 
 If you configure this option, the new Custom Field becomes visible in the list of Custom Fields. The Custom Field remains unpublished until you run the command ``bin/console mautic:custom-field:create-column``. This command creates the actual column in the table and publishes the field metadata.
 
+Similarly, the ``bin/console mautic:custom-field:delete-column`` command deletes the actual column in the table if the ``create_custom_field_in_background`` config option is set to true. The column gets soft-deleted and removed from the user interface. But the data is still present in the database until you run the command to delete the column.
```
RCheesley:
```suggestion
Similarly, the ``bin/console mautic:custom-field:delete-column`` command deletes the actual column in the table if you have turned on the ``create_custom_field_in_background`` config option. The column gets soft-deleted and removed from the user interface, but the data is still present in the database until you run the command to delete the column.
```

Passive -> Active voice

RCheesley (changes requested):
Just a small Vale fix here. Thanks for the update to the docs!

PR #353, opened 2025-03-13
Add documentation for Page draft

dadarya0:
This PR adds documentation for the Page Draft feature found in https://github.com/mautic/mautic/pull/13937


docs/components/landing_pages.rst
```diff
@@ -6,4 +6,51 @@ Landing Pages
...
+
+.. code:: php
+
+  'page_draft_enabled' => 1
+
+Once turned on, the 'Save as Draft' button appears on the Landing Page edit interface.
+
+.. image:: images/landing_pages/save-draft.png
+  :width: 400
+  :alt: Screenshot showing the 'Save as Draft' button on the landing page edit page.
```
escopecz: [vale] reported by reviewdog 🐶
[Mautic.FeatureList] Is this referring to a Mautic feature? If so, use 'Landing Page' instead of 'landing page'.

```suggestion
  :alt: Screenshot showing the 'Save as Draft' button on the Landing Page edit page.
```

docs/components/landing_pages.rst
```diff
@@ -6,4 +6,51 @@ Landing Pages
...
+
+.. code:: php
+
+  'page_draft_enabled' => 1
+
+Once turned on, the 'Save as Draft' button appears on the Landing Page edit interface.
+
+.. image:: images/landing_pages/save-draft.png
+  :width: 400
+  :alt: Screenshot showing the 'Save as Draft' button on the Landing Page edit page.
```
escopecz: @RCheesley I'm not sure how to deal with this Vale warning

> Is this referring to a Mautic feature? If so, use 'Landing Page' instead of 'page'.

It should be lower case `p`. Do we have to ignore the Vale check on this line?

RCheesley: If you're referring to a Mautic Landing Page then both words are capitalised.

RCheesley (approved):
Hi folks,

A few things which I've fixed in this PR so we can merge it:

1. You must always ensure that the headings are the same underline length as the text above them - if you run the command `make html` in your /docs folder it will flag up such warnings to you, tab complete when you start typing if you use VSCode _usually_ does it right but sometimes misses the odd character!
2. I think the heading should be H2 instead of H3 so I changed those.
3. There was some inconsistency with draft v Draft, I've matched the UI with using the capitalised version
4. An Landing Page Draft allows changes to the content of the Landing Page only. > changed to active voice: You can only change the content of the Landing Page itself when it's in Draft.
5. An Landing Page Draft may be previewed by appending > changed to active voice: You can preview a Landing Page Draft by appending
6. a Draft Preview URL will be > a Draft Preview URL is
7. There was a typo on line 49 which I fixed (val on rather than vale on) which was hiding some of the later errors.

PR #358, opened 2025-04-30
clarify messenger:consume params

O-Mutt:
If this process is setup on a cron, depending on the frequency of the cron, this will cause an OOM situation over time since the process will never exit

### DIFF
--- a/docs/configuration/cron_jobs.rst
+++ b/docs/configuration/cron_jobs.rst
@@ -116,7 +116,7 @@ If the system configuration is queueing Emails, a cron job processes them. If yo
 
 .. code-block:: php
 
-    php /path/to/mautic/bin/console messenger:consume email
+    php /path/to/mautic/bin/console messenger:consume email --time-limit=160
 
 .. vale off
 
--- a/docs/configuration/cron_jobs.rst
+++ b/docs/configuration/cron_jobs.rst
@@ -112,7 +112,7 @@ Process Email queue cron job
 
 .. vale on
 
-If the system configuration queues Emails, a Cron job processes them. If you plan to run `messenger:consume` using a Cron job, you should include at least one of these parameters: `--memory-limit`, `--limit` - the number of emails - or, `--time-limit`; otherwise, the command starts a long-lived process.
+If the system configuration queues Emails, a Cron job processes them. If you plan to run ``messenger:consume`` using a Cron job, you should include at least one of these parameters: ``--memory-limit``, ``--limit`` - the number of emails, or ``--time-limit``. This starts a long-lived process and continues to run without one of these parameters.
 
 .. code-block:: bash
 
--- a/docs/configuration/cron_jobs.rst
+++ b/docs/configuration/cron_jobs.rst
@@ -112,7 +112,7 @@ Process Email queue cron job
 
 .. vale on
 
-If the system configuration queues emails, a cron job processes them. If you plan to run `messenger:consume` using a Cron job, you should include at least one of these parameters: `--memory-limit`, `--limit` (number of emails), or `--time-limit`; otherwise, the command starts a long-lived process.
+If the system configuration queues emails, a cron job processes them. If you plan to run `messenger:consume` using a Cron job, you should include at least one of these parameters: `--memory-limit`, `--limit` - the number of emails - or, `--time-limit`; otherwise, the command starts a long-lived process.
 
 .. code-block:: php
 
--- a/docs/configuration/cron_jobs.rst
+++ b/docs/configuration/cron_jobs.rst
@@ -112,7 +112,7 @@ Process Email queue cron job
 
 .. vale on
 
-If the system configuration is queueing Emails, a cron job processes them.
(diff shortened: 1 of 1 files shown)


cibero42 (approved):

docs/configuration/cron_jobs.rst
```diff
@@ -112,11 +112,11 @@ Process Email queue cron job
 
 .. vale on
 
-If the system configuration is queueing Emails, a cron job processes them.
+If the system configuration is queueing Emails, a cron job processes them. If you plan to run `messenger:consume` on a cron you should include, at minimum one of these parameters `--memory-limit`, `--limit` (number of emails), `--time-limit` otherwise, this will start up a long-lived process.
```
RCheesley:
```suggestion
If the system configuration is queueing Emails, a cron job processes them. If you plan to run `messenger:consume` on a cron you should include, at minimum, one of these parameters `--memory-limit`, `--limit` - the number of emails - or `--time-limit`, otherwise this starts up a long-lived process.
```

RCheesley (changes requested):
Thanks for the PR @O-Mutt and welcome to the community!

A couple of small readability suggestions, plus the Vale suggestions to fix :)

docs/configuration/cron_jobs.rst
RCheesley: Would it be helpful to show how you'd use this in real life with an example command, so people understand how to specify the numbers?

O-Mutt: > Would it be helpful to show how you'd use this in real life with an example command, so people understand how to specify the numbers?

I did update the [command just below](https://github.com/mautic/user-documentation/pull/358/files#diff-84e2392b487553c90d7101e25e7ace972ea4af522f73c2d5fdbab3b191187944R119). Maybe there should be more?

docs/configuration/cron_jobs.rst
```diff
@@ -112,11 +112,11 @@ Process Email queue cron job
 
 .. vale on
 
-If the system configuration is queueing Emails, a cron job processes them.
+If the system configuration queues emails, a cron job processes them. If you plan to run `messenger:consume` using a Cron job, you should include at least one of these parameters: `--memory-limit`, `--limit` - the number of emails - or, `--time-limit`; otherwise, the command starts a long-lived process.
 
 .. code-block:: php
```
RCheesley:
```suggestion
.. code-block:: bash
```

I think this needs to be bash rather than php, perhaps.

O-Mutt:
Looks like this change passed but the build overall is broken on the messages page?

docs/configuration/cron_jobs.rst
RCheesley: That's great, I think you updated it as I wrote the comment!

docs/configuration/cron_jobs.rst
```diff
@@ -112,11 +112,11 @@ Process Email queue cron job
 
 .. vale on
 
-If the system configuration is queueing Emails, a cron job processes them.
+If the system configuration queues Emails, a Cron job processes them. If you plan to run `messenger:consume` using a Cron job, you should include at least one of these parameters: `--memory-limit`, `--limit` - the number of emails - or, `--time-limit`; otherwise, the command starts a long-lived process.
```
RCheesley:
```suggestion
If the system configuration queues Emails, a Cron job processes them. If you plan to run ``messenger:consume`` using a Cron job, you should include at least one of these parameters: ``--memory-limit``, ``--limit`` - the number of emails, or ``--time-limit``; otherwise there's a chance of the command starting a long-lived process.
```

I think this works better when reading it, see what you think. Double backticks are needed for inline code on RST, also!

RCheesley (changes requested):
Just a couple of small suggestions.

docs/configuration/cron_jobs.rst
O-Mutt: :thinking: "there's a chance" is a little less than clear in this case. If the user doesn't use one of those params it **is going to** start a long-lived process and use memory until the host goes OOM and dies

docs/configuration/cron_jobs.rst
O-Mutt: Yeah, likely correct. That was in place before I got here but happy to make it correct :)

docs/configuration/cron_jobs.rst
O-Mutt:
```suggestion
If the system configuration queues Emails, a Cron job processes them. If you plan to run ``messenger:consume`` using a Cron job, you should include at least one of these parameters: ``--memory-limit``, ``--limit`` - the number of emails, or ``--time-limit``. This starts a long-lived process and continues to run without one of these parameters.
```
WDYT?

RCheesley: Ah OK, sounds good!

RCheesley: Did you want to go ahead and resolve the conversations, then we can get this merged.

RCheesley (approved):

PR #361, opened 2025-05-09
Campaign import export user docs

dcjarvis:
Added descriptions for campaign import and campaign export.


dcjarvis:
Note that these docs should be targeted to Mautic v7.x but that branch is not available yet.

docs/some-dir/README.md
dcjarvis, line 1: File was committed in error

docs/some-dir/composer.json
dcjarvis, line 1: File was committed in error

docs/campaigns/exporting_campaigns.rst
dcjarvis, line 1: This is an exception because we are referring to a Mautic feature in the title.

dcjarvis:
Added further parameter to the command line for export

docs/configuration/command_line_interface.rst
```diff
@@ -115,14 +115,21 @@ These are the commands you may need to use in relation to your Mautic instance.
...
+     - Imports campaign and dependent entities to Mautic from a ZIP file. See :doc:`/campaigns/importing-campaigns`.
+   
+   * - ``mautic:entity:export --entity=campaign --id=1 path=path/to-file``
+     - Exports campaign and dependent entities from Mautic to a ZIP file. See :doc:`/campaigns/exporting-campaigns`.
+   
    * - ``messenger:consume email``
      - Processes mail queue
      - 
-   * - ``mautic:fields:analse``
+   * - ``mautic:fields:analyse``
```
RCheesley, line 131: Given we're en_US should we be using mautic:fields:analyze here @escopecz ?

RCheesley (commented):
I'm not sure if we should be using the English spelling in this cron job - I think it should be mautic:fields:analyze - @escopecz is this something we can change easily in the code? I am sure it will cause confusion in our users using the British English version.

docs/configuration/command_line_interface.rst
escopecz: Mautic is using US English if I'm not mistaken. It can be changed in the code quite easily. And in fact it's not that hard to support both. One will be an alias of the other.

adiati98: If one will be an alias of the other, can we use `mautic:fields:analyze` here, as our docs based on en_US?

escopecz: Sorry, such alias isn't present. We'll have to stick with `s`

RCheesley: OK, something to consider for future reference. I can see this tripping people up in the future because they're so used to USA language but it's done now!

docs/campaigns/images/campaign-sample.json
adiati98, line 1: I don't see we use this file in any content. Do we still need this?

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,155 @@
...
+* Other related dependencies
+
+The export command:
+
+* Detect use of Plugins and Custom Fields
+* Include the data to support these dependencies
+
+.. important::
+    
+   The importing instance needs the same Custom Fields and Plugins to be present.
```
adiati98, line 30: I'm not sure if user docs need these info.

But if we do need it, I particularly don't understand about the export command here. What do we actually want to explain?

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,155 @@
...
+Data structure
+    Exports data in a structured JSON format to ensure compatibility.
+
+Asset management
+    Exports Assets into a separate folder in their original format.
+
+File packaging
+    Zips the resulting collection of files into a single package for easy transfer across systems.
+
+.. vale off
```
adiati98: I'm thinking if we should move this section to after the "Exporting a Campaign" section. Because in my opinion, this is more like a supplement info if users want to know how the feature works under the hood.

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,155 @@
...
+
+You can programmatically export Campaigns using the Mautic API. To do this, you must authenticate your API request with the credentials stored in Mautic's settings.
+
+.. vale off
+
+For more details on how to authenticate, see the :doc:`Authentication page </authentication/authentication>`.
+
+.. vale on
+
+cURL example
```
adiati98, line 106: Do we want to use "Curl", "cURL", or "curl"?

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,230 @@
...
+    * Update existing entities, allowing Administrators to update existing Campaigns.
+    * Create new entities using a new ID.
+
+Automatic data mapping
+    Mautic maps imported data to the correct locations and automatically creates any necessary dependent entities.
+
+Campaign activation
+    After a successful import, the Campaign remains inactive by default.
+
+.. vale off
```
adiati98: Same comment here with the one in the `exporting_campaigns.rst` regarding moving this section down.

adiati98 (commented):
@mautic/core-team @RCheesley,

As these are new pages, I deliberately took the initiative to adjust the wordings, restructure, and add images to make them clearer and consistent.

Therefore, I need your help to review the content thoroughly.

Additionally, I left some questions in this PR on the comments. TIA! ✨

adiati98:
> @mautic/core-team @RCheesley,
> 
> As these are new pages, I deliberately took the initiative to adjust the wordings, restructure, and add images to make them clearer and consistent.
> 
> Therefore, I need your help to review the content thoroughly.
> 
> Additionally, I left some questions in this PR on the comments. TIA! ✨

@mautic/core-team any thoughts on this?

docs/campaigns/images/campaign-sample.json
escopecz: It can be removed if unused.

docs/campaigns/exporting_campaigns.rst
escopecz: @vukovicpredrag can you validate? In the paragraph above it states that custom fields are part of the extraction and in this sentece that it needs custom fields must be present. I'm unsure which is true.

vukovicpredrag: @escopecz @adiati98  Custom Fields are not included in the campaign export. The export only contains references to them. The importing instance needs the same custom fields already present.

docs/campaigns/exporting_campaigns.rst
vukovicpredrag: In my opinion, this section should be moved after the "Exporting a Campaign" section. Practical steps first, technical details second.

docs/campaigns/exporting_campaigns.rst
vukovicpredrag: cURL - that's the official spelling.

docs/campaigns/importing_campaigns.rst
vukovicpredrag: Same as my comment on the export page - move this after the practical import steps.

docs/campaigns/exporting_campaigns.rst
adiati98: @vukovicpredrag @escopecz now I'm confused. 😅

Which one should we omit and which to keep?

escopecz: I think Like this is correct for the current state:
Screenshot 2026-03-06 at 15 59 25

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
+Exporting Campaigns
+###################
+
+Before importing or exporting data, back up your database as a safety precaution in case you need to restore it. Speak to the administrator of your domain to do this, as it requires specific technical knowledge.
```
RCheesley:
```suggestion
Before importing or exporting data, back up your database as a safety precaution in case you need to restore it. Speak to the administrator of your Mautic instance to do this, as it requires specific technical knowledge.
```

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+
+   Both the import and export features are available only in Mautic 7.0 and later versions.
+
+Supported data types
+********************
+
+When you select a Campaign, the export feature extracts all Campaign data and entities, along with any dependencies the Campaign needs to function, including:
+
+* Dynamic Content
+* Asset\*
```
RCheesley, line 16: What does this asterisk indicate? I'd like us to include the reference if we're referring to something, right after these bullets.

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+For more details on how to authenticate, see the :doc:`Authentication page </authentication/authentication>`.
+
+.. vale on
+
+cURL example
+------------
+
+.. code-block:: bash
+
+   curl --location 'https://{your-mautic-domain}/api/campaigns/export/1' \
```
RCheesley: I wonder if we should make it explicit here that you replace everything in {your-mautic-domain} including the curly braces, or would that be obvious? Also we use that here, and example.com later. Why not use the same?

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,224 @@
...
+
+   |
+
+   .. tip::
+
+      \* Use a ZIP file created from the Mautic export function - **recommended**
+
+#. Ensures inclusion of Campaign data, external Assets, and Dynamic Content.
+
+#. Select **Actions** option from the dropdown menu for the **Campaign** and **Segment** entities. Choose either **Update entity** or **Create new entity**.
```
RCheesley, line 43: I think we should explain what each does and why you see it - e.g. update entity suggests that it has found this already existing in Mautic, and it'll update with any changes compared to what's there (could be destructive, might overwrite changes). Create new will force it to create a new entity, even if something that matches is found. Safer option but can lead to duplication of resources.

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,224 @@
...
+#. Click the red toggle button next to the Campaign's name to change the status to active.
+
+   |
+
+   .. image:: images/campaign_inactive_toggle_button.png
+      :alt: Highlight of a Campaign's inactive toggle button
+
+   |
+
+#. Click **Yes** when a prompt message appears.
```
RCheesley, line 92: We should link to the docs which explain what this means. Clicking yes can have consequences. Publishing a campaign without checking it first could also be disastrous. I think we need to add a clear warning here, that once you publish a campaign, it immediately starts running. People must check that they understand what the campaign is doing, and verify each step of the way, before they decide to turn it on. Red warning admonition called for, I think.

RCheesley (changes requested):
A few small suggestions!

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
```
adiati98:
```suggestion
   curl --location 'https://example.com/api/campaigns/export/1' \
```

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,221 @@
...
+
+   .. image:: images/upload_zip_file.png
+      :alt: The Choose file and Upload buttons to import Campaigns
+
+   |
+
+   .. tip::
+
+      \* Use a ZIP file created from the Mautic export function - **recommended**. This ensures inclusion of Campaign data, external Assets, and Dynamic Content.
+#. Select **Actions** option from the dropdown menu for the **Campaign** and **Segment** entities. Choose either **Update entity** or **Create new entity**.
```
adiati98, line 43:
```suggestion
#. Select **Actions** option from the dropdown menu for the **Campaign** and **Segment** entities. Choose either **Update entity** or **Create new entity**.

   These options appear when the import process detects existing Campaigns or Segments that match the imported data.

   * **Update entity** - Overwrites the existing Campaign or Segment in Mautic with the imported data. Use this option to update configuration, but note that it can overwrite or destroy existing changes on your instance.
   * **Create new entity** - Forces Mautic to create a brand-new Campaign or Segment, even if a matching one already exists. This is the safer option to prevent data loss, but it can lead to duplicate resources.
```

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,221 @@
...
+#. Click the red toggle button next to the Campaign's name to change the status to active.
+
+   |
+
+   .. image:: images/campaign_inactive_toggle_button.png
+      :alt: Highlight of a Campaign's inactive toggle button
+
+   |
+
+#. Click **Yes** when a prompt message appears.
```
adiati98, line 92:
```suggestion
#. Click **Yes** when a prompt message appears.

   .. warning::
          
   Activating a Campaign starts execution immediately. Review all Campaign steps, configurations, and associated Assets thoroughly before turning on a Campaign to prevent unintended execution of events. 
          
   When a Campaign goes live, scheduled events with relative delays follow the **Campaign Republish Behavior** configuration. If the Campaign uses a specific setting instead of the global default, the execution behavior changes:

   * **Use global setting** - Applies the default behavior configured in the global Mautic settings.
   * **Restart on republish** - Resets the delay timer completely. The delay period starts over from zero the moment the Campaign becomes active.
   * **Count delay only while published** - Pauses the delay timer while the Campaign is inactive. The timer resumes from where it paused once the Campaign becomes active again.
   * **Count delay regardless of publish state** - default option. Keeps the delay timer running continuously in the background, even while the Campaign remains inactive.
```

adiati98 (changes requested):
Let's address the review from @RCheesley.

adiati98:
@mautic/core-team can you please review the suggested changes by me here if they're correct in addressing @RCheesley's reviews? TIA! ✨

RCheesley:
Thank you for reaching out. I am currently unavailable, but I have alerted the team who will respond as appropriate.

docs/campaigns/exporting_campaigns.rst
vukovicpredrag: Standardizing on example.com works and resolves the inconsistency Ruth flagged. One thing: a bare example.com doesn't signal "replace this" as clearly as a {placeholder} would, so let's make sure the surrounding text tell  the reader to swap example.com for their own Mautic instance domain. With that note added, this looks good.

docs/campaigns/importing_campaigns.rst
vukovicpredrag: Descriptions are accurate and match Ruth's request. Optional: a short lead-in saying these options appear because the import found a matching entity would make the "why you see it" part explicit. Otherwise looks good.

docs/campaigns/importing_campaigns.rst
vukovicpredrag: The warning addresses Ruth's point well. Two fixes on the option list:

  1. Typo: "Count delay on while published" - should be "Count delay only while published" (the description is correct). 
  2. "Count delay regardless of publish state" is the default option - worth stating that explicitly.

Non-blocking: the Republish Behavior detail applies to republishing in general rather than to import specifically it could live in a note or a link to keep the import step focused.

docs/campaigns/exporting_campaigns.rst
adiati98: @vukovicpredrag we actually have an "important" admonition to address replacing the placeholder at the bottom:

https://github.com/dcjarvis/mauticuser-documentation/blob/b61e63848885b8e42cb528b0889d742f662ffd2c/docs/campaigns/exporting_campaigns.rst?plain=1#L125-L130 

Now that you mentioned that, this might get missed if folks don't read the full docs until the next example. I'll see what we can do to make this noticable. 👍🏼

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+   --header 'Authorization: Bearer YOUR_ACCESS_TOKEN' \
+   --data ''
```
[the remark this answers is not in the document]
adiati98: @vukovicpredrag is this review from co-pilot correct? That we should remove the `--data ''`?

docs/campaigns/importing_campaigns.rst
adiati98: Does the short lead-in here clear enough?

vukovicpredrag:
Thanks @adiati98 - all four points look good from my side (Copilot's right on --data '' forcing POST, the lead-in reads clearly, Republish Behavior fixes are in, and surfacing the existing placeholder admonition covers Ruth's point). No blockers from me!

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+
+3.\ Using Mautic API
+====================
+
+You can programmatically export Campaigns using the Mautic API. To do this, you must authenticate your API request with the credentials stored in Mautic's settings.
+
+.. vale off
+
+For more details on how to authenticate, see the :doc:`Authentication page </authentication/authentication>`.
+
```
adiati98, line 93:
```suggestion

.. important::

   * The API uses a GET request to export a specific Campaign by ID.   
    
   * Before running the following cURL or Python examples, ensure you:

     * Replace ``example.com`` with the actual Mautic instance domain.
     * Replace ``YOUR_ACCESS_TOKEN`` with a valid authentication token.
     * Verify that the User account has the necessary API permissions.

```

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+       export_file = response.content
+       # Save or process the exported campaign
+
+.. important::
+
+   * Replace ``example.com`` with your actual Mautic instance domain
+   * Replace ``YOUR_ACCESS_TOKEN`` with a valid authentication token
+   * The API uses a GET request to export a specific Campaign by ID
+   * Ensure you have the necessary API permissions
+
```
adiati98:
```suggestion
```

adiati98 (changes requested):
Move API replacement instructions before code examples for better visibility.

docs/campaigns/exporting_campaigns.rst
adiati98: @vukovicpredrag I think moving the instructions to before the code examples like this would be better for visibility?

docs/campaigns/importing_campaigns.rst
adiati98: > Non-blocking: the Republish Behavior detail applies to republishing in general rather than to import specifically it could live in a note or a link to keep the import step focused.

I think that's a great idea as I don't think we have a dedicated section or whatsoever in the docs for this info. I'll create an issue to address this. 👍🏼

docs/campaigns/exporting_campaigns.rst
vukovicpredrag: @adiati98 Moving it above the code blocks (or even repeating a short "replace example.com" note inline) should solve the visibility issue. Either way works for me readers  just need to see it before they copy-paste the curl.

docs/campaigns/exporting_campaigns.rst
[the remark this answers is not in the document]
vukovicpredrag: Yes, Copilot is right. --data flag forces curl to POST even with an empty body, so dropping --data '' makes it a proper GET (which is what the export endpoint  expects). Apply the suggestion.

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+Supported data types
+********************
+
+When you select a Campaign, the export feature extracts all Campaign data and entities, along with any dependencies the Campaign needs to function, including:
+
+* Dynamic Content
+* Asset\*
+* Other related dependencies
+
```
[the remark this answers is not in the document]
vukovicpredrag: Copilot's suggested footnote ("The exported Campaign includes only directly associated Assets") is accurate and clears up what the asterisk refers to. Apply the suggestion.

docs/campaigns/importing_campaigns.rst
vukovicpredrag: Yes, "These options appear when the import process detects existing Campaigns or Segments that match the imported data" nails the "why am I seeing this" part. Reads well  good to go. 👍

docs/campaigns/importing_campaigns.rst
vukovicpredrag: Sounds good separate issue for the Republish Behavior section makes sense. We can keep the warning here focused on activation, then link out once the dedicated  section exists.

docs/campaigns/exporting_campaigns.rst
vukovicpredrag: Yes, above the code examples is the right.

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,149 @@
...
+* ``--entity``: defines the type of entity to export, in this case, ``campaign``.
+* ``--id``: defines the ID of the Campaign to export. When you view or edit the Campaign, look at the URL to find the ID. For example, ``/s/campaigns/view/123``, where ``123`` is the ID.
+* ``--zip-file``: creates a ZIP file of the Campaign and its dependencies.
+* ``--path``: specifies the directory to save the exported file.
+
+**2. In a JSON file**
+
+.. code-block:: bash
+    
+    bin/console mautic:entity:export --entity=campaign --id=1 --json-file
```
vukovicpredrag: One technical correction - in the "In a JSON file" example, the flag is --json-only (not --json-file). Verified against EntityExportCommand in core. Otherwise the  CLI/API examples all match the codebase.

vukovicpredrag:
Went through the full docs - overall looks good. Left a few inline comments on small technical fixes, rest is solid. Good job.

docs/configuration/command_line_interface.rst
```diff
@@ -115,15 +118,22 @@ These are the commands you may need to use in relation to your Mautic instance.
...
      - Analyze Custom Fields table and return table or file with results. See :doc:`/contacts/custom_fields`.
      - 
    * - ``mautic:import``
-     - If the CSV import is configured to run in background then this command will pick up the pending import jobs and imports the data from CSV files to Mautic.
+     - | Imports Contacts from a CSV file
+       | If the CSV import is configured to run in background then this command will pick up the pending import jobs and imports the data from CSV files to Mautic.
      - 
    * - ``mautic:integration:fetchleads``
      - Fetch Contacts from Integration.
```
vukovicpredrag, line 139: The --zip-file flag in EntityExportCommand is defined as VALUE_NONE, meaning it acts as a boolean switch and does not accept a value. The current example will fail  when copy-pasted because of the =path/to-file/campaign_export.zip part.
  
Correct form:

``mautic:entity:export --entity=campaign --id=1 --zip-file --path=path/to-file``

docs/campaigns/importing_campaigns.rst
adiati98: @vukovicpredrag quick questions:

- Does Republish behavior only used in Campaign? If not, where else?
- Do you happen to know when did this first introduce? M7 or M5? I need the info for backport matter.

TIA!

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,221 @@
...
+    
+   * Replace ``example.com`` with your actual Mautic instance domain.
+   * Ensure you have a valid access token by accessing the API Credentials section within Mautic's settings.
+   * The imported Campaign must comply with Mautic's Campaign structure.
+   * Verify import permissions and data integrity.
+
+.. vale off
+
+How Campaign import works
+********************************
```
adiati98:
```suggestion
How Campaign import works
*************************
```

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,237 @@
...
+
+   .. tip::
+
+      \* Use a ZIP file created from the Mautic export function - **recommended**. This ensures inclusion of Campaign data, external Assets, and Dynamic Content.
+#. Select **Actions** option from the dropdown menu for the **Campaign** and **Segment** entities. Choose either **Update entity** or **Create new entity**.
+
+   These options appear when the import process detects existing Campaigns or Segments that match the imported data.
+
+   * **Update entity** - Overwrites the existing Campaign or Segment in Mautic with the imported data. Use this option to update configuration, but note that it can overwrite or destroy existing changes on your instance.
+   * **Create new entity** - Forces Mautic to create a brand-new Campaign or Segment, even if a matching one already exists. This is the safer option to prevent data loss, but it can lead to duplicate resources.
```
adiati98, line 50: The word 'lead' that Vale warns is a verb. So, we can safely wrap this in vale off/on statements.

```suggestion
.. vale off

   * **Update entity** - Overwrites the existing Campaign or Segment in Mautic with the imported data. Use this option to update configuration, but note that it can overwrite or destroy existing changes on your instance.
   * **Create new entity** - Forces Mautic to create a brand-new Campaign or Segment, even if a matching one already exists. This is the safer option to prevent data loss, but it can lead to duplicate resources.

.. vale on
```

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,237 @@
...
+   .. warning::
+          
+   Activating a Campaign starts execution immediately. Review all Campaign steps, configurations, and associated Assets thoroughly before turning on a Campaign to prevent unintended execution of events. 
+          
+   When a Campaign goes live, scheduled events with relative delays follow the **Campaign Republish Behavior** configuration. If the Campaign uses a specific setting instead of the global default, the execution behavior changes:
+
+   * **Use global setting** - Applies the default behavior configured in the global Mautic settings.
+   * **Restart on republish** - Resets the delay timer completely. The delay period starts over from zero the moment the Campaign becomes active.
+   * **Count delay only while published** - Pauses the delay timer while the Campaign is inactive. The timer resumes from where it paused once the Campaign becomes active again.
+   * **Count delay regardless of publish state** - default option. Keeps the delay timer running continuously in the background, even while the Campaign remains inactive.
```
adiati98, line 105: The word 'published' that Vale flags here are what appears on the UI. Let's wrap this in vale off/on statements.


```suggestion
   .. vale off
   
   * **Use global setting** - Applies the default behavior configured in the global Mautic settings.
   * **Restart on republish** - Resets the delay timer completely. The delay period starts over from zero the moment the Campaign becomes active.
   * **Count delay only while published** - Pauses the delay timer while the Campaign is inactive. The timer resumes from where it paused once the Campaign becomes active again.
   * **Count delay regardless of publish state** - default option. Keeps the delay timer running continuously in the background, even while the Campaign remains inactive.
   
   .. vale on
```

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,154 @@
...
+
+   # Handle response
+   if response.status_code == 200:
+       export_file = response.content
+       # Save or process the exported campaign
+
+.. vale off
+
+How Campaign export works
+*********************************
```
adiati98: Remove excess asterisks.


```suggestion
How Campaign export works
*************************
```

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,154 @@
...
+* Asset\*
+* Other related dependencies
+
+.. note::
+
+   \* The exported Campaign includes only directly associated Assets.
+The export command:
+
+* Detect use of Plugins and Custom Fields
+* Include the data to support these dependencies
```
adiati98, line 26: Add a blank line.


```suggestion

The export command:

* Detect use of Plugins and Custom Fields
* Include the data to support these dependencies
```

adiati98 (commented):
We need to address the Vale warnings and tidy up some things.

vukovicpredrag (commented):
Two small CLI flag fixes - details inline:
  
 1. exporting_campaigns.rst: `--json-file` should be `--json-only` (that flag doesn't exist).
 2. command_line_interface.rst: `--zip-file` is a boolean (VALUE_NONE), so `--zip-file=path/...` fails. Use `--zip-file --path=path/to-file`.
  
Verified against EntityExportCommand / EntityImportCommand in core. Rest of the examples match the codebase.

docs/campaigns/exporting_campaigns.rst
```diff
@@ -0,0 +1,155 @@
...
+* ``--entity``: defines the type of entity to export, in this case, ``campaign``.
+* ``--id``: defines the ID of the Campaign to export. When you view or edit the Campaign, look at the URL to find the ID. For example, ``/s/campaigns/view/123``, where ``123`` is the ID.
+* ``--zip-file``: creates a ZIP file of the Campaign and its dependencies.
+* ``--path``: specifies the directory to save the exported file.
+
+**2. In a JSON file**
+
+.. code-block:: bash
+    
+    bin/console mautic:entity:export --entity=campaign --id=1 --json-file
```
adiati98:
```suggestion
    bin/console mautic:entity:export --entity=campaign --id=1 --json-only
```

docs/configuration/command_line_interface.rst
```diff
@@ -115,15 +118,22 @@ These are the commands you may need to use in relation to your Mautic instance.
      - 
    * - ``mautic:email:fetch``
      - Fetch and process monitored Email.
-     - 
+     -
+   * - ``mautic:entity:import --entity=campaign --file=path-to-file/entity_data.zip --user=1``
+     - Imports Campaign and dependent entities to Mautic from a ZIP file. See :doc:`/campaigns/importing_campaigns`.
+     -
+   * - ``mautic:entity:export --entity=campaign --id=1 --zip-file=path/to-file/campaign_export.zip --path=path/to-file``
```
adiati98:
```suggestion
   * - ``mautic:entity:export --entity=campaign --id=1 --zip-file --path=path/to-file``
```

adiati98 (commented):
@vukovicpredrag sorry, I missed your latest review! Are these changes addressing your review correctly?

vukovicpredrag:
@adiati98 No worries. Both suggestions are correct and address the review exactly. Thank you!

adiati98:
> @adiati98 No worries. Both suggestions are correct and address the review exactly. Thank you!

I've committed the changes. Can you please review this again or approve it if you feel like everything's already good to go? Thanks, @vukovicpredrag!

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,245 @@
...
+      url = 'https://example.com/api/campaigns/import'
+
+      # Authentication
+      headers = {
+          'Authorization': 'Bearer YOUR_ACCESS_TOKEN',
+          'Content-Type': 'application/json'
+      }
+
+      # Campaign import data
+      payload = {
```
vukovicpredrag: This JSON import example won't work as written. `CampaignApiController::importCampaignAction` requires the request body to be the full campaign export structure (it  checks `isset($data[0]['campaign'])`). A flat `{ "name": ..., "description": ... }` payload fails that check and returns `400`. The endpoint also returns `201`, not  `200`.

The body must be the same JSON as `entity_data.json` from an export ZIP (or the response of `GET /api/campaigns/export/{id}`). 

Suggested fix:

  ```python

  import requests
  import json

  url = 'https://example.com/api/campaigns/import'

  headers = {
      'Authorization': 'Bearer YOUR_ACCESS_TOKEN',
      'Content-Type': 'application/json'
  }

  # Body must be the full campaign export structure (same as entity_data.json)
  with open('entity_data.json') as f:
      payload = json.load(f)
  
  response = requests.post(url, headers=headers, json=payload)

  if response.status_code == 201:
      print("Campaign imported successfully")
  ```

vukovicpredrag (approved):
Re-reviewed against core - everything checks out (UI/CLI import, export flags, API routes, Republish Behavior, plugin validation). Only one minor note on the JSON  import example (inline) - can be a quick follow-up. Approving! Thanks @adiati98 🚀

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,245 @@
...
+          # Add other campaign details as needed
+      }
+
+      # Send import request
+      response = requests.post(url, headers=headers, json=payload)
+
+      # Handle response
+      if response.status_code == 200:
+          imported_campaign = response.json()
+          print("Campaign imported successfully")
```
adiati98, line 195:
```suggestion
   .. code-block:: python

      import requests
      import json

      # API Endpoint
      url = 'https://example.com/api/campaigns/import'

      # Authentication
      headers = {
          'Authorization': 'Bearer YOUR_ACCESS_TOKEN',
          'Content-Type': 'application/json'
      }

      # Campaign import data
      with open('entity_data.json') as f:
          payload = json.load(f)
      }

      # Send import request
      response = requests.post(url, headers=headers, json=payload)

      # Handle response
      if response.status_code == 201:
          print("Campaign imported successfully")
```

docs/campaigns/importing_campaigns.rst
adiati98: @vukovicpredrag I've applied the suggestion by copy-pasting the code block. I guess everything's good to go?

docs/campaigns/importing_campaigns.rst
```diff
@@ -0,0 +1,243 @@
...
+      # API Endpoint
+      url = 'https://example.com/api/campaigns/import'
+
+      # Authentication
+      headers = {
+          'Authorization': 'Bearer YOUR_ACCESS_TOKEN',
+          'Content-Type': 'application/json'
+      }
+
+      # Campaign import data
```
adiati98: Follow the comment in the suggestion.

```suggestion
      # Body must be the full campaign export structure (same as entity_data.json)
```

docs/campaigns/importing_campaigns.rst
vukovicpredrag: Sorry for the late reply @adiati98 , this slipped past me.  I've gone through the latest changes (including the Python example fix) and they all look correct  nothing further from my side, my approval stands. Thanks for your patience!

adiati98: Thanks so much for your help and review, @vukovicpredrag!

RCheesley (approved):
LGTM, thanks everyone for the great work!

adiati98 (approved):
Thank you, everyone!

adiati98:
@all-contributors please add @dcjarvis for docs

adiati98:
@all-contributors please add @vukovicpredrag for review

PR #362, opened 2025-05-22
Add support for GitHub Codespaces

matbcvo:
Duplicate of PR https://github.com/mautic/developer-documentation-new/pull/246

image

Wait until the "Setting up remote connection: Building codespace..." has finished.

image

Once the codespace has been built, a terminal will open and run some post-creation commands (installing required Python packages and Vale). This step is quick.

Next, open any .rst file and click the Preview button at the top right.

image

The preview window will open, but it won’t work. [See here](https://docs.esbon.io/en/latest/lsp/getting-started.html#open-a-preview)
There’s a workaround: opening the preview triggers the Esbonio server to start. Then, under the Ports tab, you’ll see a URL that you can use to view the documentation. Click either of the two buttons - the first opens it in the browser, the second opens it in the editor.

image

When viewing via the URL, you’ll need to refresh the page after making edits.

https://github.com/user-attachments/assets/8bee11ad-15d1-47bc-821f-34534fd17cc6

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+{
+    "features": {
+        "ghcr.io/devcontainers/features/python:1": {
+            "version": "3.11",
+            "installTools": true
+        }
+    },
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "ms-python.python",
+                "swyddfa.esbonio",
+                "lextudio.restructuredtext",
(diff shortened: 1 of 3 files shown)


RCheesley (approved):
Thanks @matbcvo - this is great!

matbcvo:
@RCheesley, I’m unable to merge this PR because the linting test is failing.

Issue #399, opened 2025-06-10
Remove harmful command


adiati98:
## Description

Based on [this comment](https://github.com/mautic/mautic/issues/15081#issuecomment-2955751016), we need to remove the `bin/console doctrine:schema:update --no-interaction --force` from the docs in all versions as it's not necessary and quite harmful.

## Proposed Solution

As we need to remove the command from several branches, there should one PR for each branch with the below list:

- [ ] 7.0
- [x] 6.0
- [x] 5.2
- [x] 5.x
- [x] 4.x
- [ ] 3.x
- [ ] 2.x

RCheesley:
I think we can focus just on the ones which are visible in the front-end of the docs, rather than all branches.

adiati98:
Noted. So I'll prioritize those in `5.2`, `6.0`, `5.x`, and `4.x`.

adiati98:
@RCheesley I've created the PRs for the prioritized branches.

FYI, I'm not sure why all checks are failed in the `4.x` branch. There's only 1 check that failed in others.

Please let me know if you'd like me to remove the command from other branches as well. Thank you.

RCheesley:
Merged all the PRs so this one should be good to close as completed - thanks @adiati98 and @escopecz for flagging it up!

PR #403, opened 2025-06-10
Remove update command from 4.x docs

adiati98:
## Description

This PR removes the `bin/console doctrine:schema:update --no-interaction --force` from `4.x` docs with link below:

https://docs.mautic.org/en/4.x/getting_started/how_to_install_mautic.html#updating-mautic-core

## Linked Issue

Relates to #399


RCheesley:
This PR is based on the 4.x branch so it won't pass any of the checks, I think.

adiati98:
If the packages have been changed and updated after `4.x`, then the setup might be different, I think.

The change in this PR is only removing the command. So, I guess it won't break anything.

RCheesley (approved):
This is GTG, the checks will fail as this branch is super old and outdated and predates us using them.

PR #349, opened 2025-02-07
feat: Add documentation for new segment building status

driskell:
Docs for https://github.com/mautic/mautic/pull/14431

Might need a `6.x` branch target for this.

PR #614, opened 2026-01-25
Cherry-pick: "feat: Add documentation for new segment building status " to 6.0

adiati98:
## Description

This PR cherry-picks https://github.com/mautic/user-documentation/pull/349 into the 6.0 branch.

Referring to the `mautic/mautic` PR:

- https://github.com/mautic/mautic/pull/14431


docs/segments/manage_segments.rst
adiati98, line 59: This section needs to be updated for the following reasons:

* There's no need to mention about viewing Contacts in Segments as this has been addressed in the first section.
* We need to add the missing info about how to export Contacts in a Segment, which is what this section is all about

docs/segments/manage_segments.rst
```diff
@@ -1,19 +1,19 @@
...
+When viewing all Segments, the **# contacts** column shows the number of Contacts included in each specific Segment. The **Building** or **Building (X Contacts)** label appears during the creation of a new Segment or when modifying an existing Segment's filter and remains visible until the building process completes.
+
+.. vale on
+
+.. image:: images/contact_segments.png
+   :width: 700
+   :alt: Highlight Segments' Contacts column that consists of Contacts count and building labels
 
 .. note::
   Mautic renamed ``lead lists`` to ``Segments`` in Mautic :xref:`1.4.0 release`.
```
adiati98: As this already M6, do we still need this info from M1?

docs/segments/manage_segments.rst
adiati98: As this already M6, do we still need this info?

adiati98 (commented):
@mautic/core-team @RCheesley,

I've left some comments here and would love your input.

We might want to update this page to ensure that the contents are still up-to-date and to make the flow clearer.

adiati98:
@RCheesley FYI, I only addressed Vale and wordings for parts that are updated in the original PR, and not the whole page. 

My concern is doing so will alter many lines of code that are not intended for the original PR. We can always do that in a new PR.

docs/segments/manage_segments.rst
escopecz: Let's remove it. Users don't need to know this. Developers do, but that's a different repo.

docs/segments/manage_segments.rst
escopecz: No. Let's take it as a fact. We don't need these version comments as we have versioned docs now for each M version. Nobody should be running M3

adiati98: I think we better do this in another PR so that this PR won't go out of track from the initial prurpose.

RCheesley: Pls add an issue to track :)

RCheesley (approved):

docs/segments/manage_segments.rst
adiati98: Created an issue and sub-issues here: https://github.com/mautic/user-documentation/issues/636

PR #310, opened 2024-06-14
Added Tip to Variables for using Custom Company Fields

LeonOltmanns:
The documentation lacks a hint, that the syntax for using core company fields as tokens and custom created onces is different.

I added a small "Tip" to briefly explain this difference.

PR #627, opened 2026-01-28
Cherry-pick: "Added Tip to Variables for using Custom Company Fields" to 5.x

adiati98:
## Description

This PR cherry-picks https://github.com/mautic/user-documentation/pull/310 into the 5.x branch as **it's made directly on the 5.x branch** and we can't co-author the PR.

This PR also addresses the build fails by fixing the external links and renames some files following the existing branches for consistency.

### DIFF
--- a/docs/configuration/variables.rst
+++ b/docs/configuration/variables.rst
@@ -137,6 +137,11 @@ See :doc:`Companies</companies/companies_overview>` for more information.
    * - Zip Code (Company)
      - ``{contactfield=companyzipcode}``
 
+.. tip:: 
+  Custom Company Fields
+
+  The syntax for custom Company fields is different than the core Company field syntax. You must **not** add the word "Company" in the variable, but instead treat it like a "contactfield":
+
 Mautic Component tokens
 ***********************
 
--- a/docs/configuration/variables.rst
+++ b/docs/configuration/variables.rst
@@ -137,9 +137,11 @@ See :doc:`Companies</companies/companies_overview>` for more information.
    * - Zip Code (Company)
      - ``{contactfield=companyzipcode}``
 
-.. tip:: Custom Company fields
+.. tip::
+  
+   **Custom Company fields**
 
-   The syntax for custom Company fields is different than the core Company field syntax. You must **not** add the word "Company" in the variable; instead, treat it like a ``contactfield``.
+   The syntax for custom Company fields is different than the core Company field syntax. You must **not** add the word "Company" in the variable; instead, treat it as a ``contactfield``.
 
 Mautic Component tokens
 ***********************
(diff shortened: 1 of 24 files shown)


docs/links/mautic_reports_api_docs.py
```diff
@@ -2,6 +2,6 @@
 
 link_name = "Reports API documentation" 
 link_text = "Reports API documentation" 
-link_url = "https://developer.mautic.org/#reports" 
+link_url = "https://devdocs.mautic.org/en/5.x/components/reports.html"
```
adiati98: The URL is incorrect.


```suggestion
link_url = "https://devdocs.mautic.org/en/5.x/rest_api/reports.html"
```

docs/configuration/variables.rst
```diff
@@ -137,6 +137,12 @@ See :doc:`Companies</companies/companies_overview>` for more information.
    * - Zip Code (Company)
      - ``{contactfield=companyzipcode}``
 
+.. tip::
+  
+   **Custom Company fields**
+
+   The syntax for custom Company fields is different than the core Company field syntax. You must **not** add the word "Company" in the variable; instead, treat it as a ``contactfield``.
```
adiati98:
```suggestion
   The syntax for custom Company fields differs from core Company field syntax. You must **not** add the word "Company" in the variable and instead treat it as a ``contactfield``.
```

docs/configuration/variables.rst
```diff
@@ -137,6 +137,12 @@ See :doc:`Companies</companies/companies_overview>` for more information.
    * - Zip Code (Company)
      - ``{contactfield=companyzipcode}``
 
+.. tip::
+  
+   **Custom Company fields**
+
+   The syntax for custom Company fields differs from core Company field syntax. You must **not** add the word "Company" in the variable and instead treat it as a ``contactfield``.
```
adiati98: @RCheesley I'm trying to keep the original version (with slight wording changes). But do you think the below is clearer and hold the same meaning? If so, I'll fix this and other PRs for other branches.

 
```suggestion
   The syntax for custom Company fields differs from core Company field syntax. You must **not** add the word 'Company' in the variable and instead treat it as a ``contactfield``.
```

RCheesley: Yes, I think this is fine. I don't know if we need to include it but it can't hurt. I wonder if it might make it less confusing to show how you refer to company fields when we say it something like this:

Unlike core Company fields which you refer to thus:

<example>

Custom fields based on the Company require the ``contactfield`` syntax, thus:

<example>

Be sure not to include the word ``company`` in the variable name.

adiati98: @RCheesley any thought on this?

RCheesley (approved):

PR #746, opened 2026-06-05
chore: Update `.all-contributorsrc` file to preserve all contributors

adiati98:
## Description

We lost many existing contributors from `.all-contributorsrc`. This PR brings back the complete list of contributors so when a new contributor is added by the bot, it won't remove contributors that are not listed in the `.all-contributorsrc` from the README.

We don't need to update the README because it somehow still shows the complete contributors.


adiati98:
Prose failures here are all contributor names. So, this is good to go.

PR #841, opened 2026-07-14
Cherry pick: "Docs for https://github.com/mautic/mautic/pull/16123" to 7.2

adiati98:
## Description

This PR cherry-picks https://github.com/mautic/user-documentation/pull/714 into the 7.2 branch.

Content is approved by @escopecz in the closed PR.

## Linked issue

https://github.com/mautic/mautic/pull/16123

PR #842, opened 2026-07-14
Fix: Table overflow from unbreakable inline code and stale static assets in docs build

adiati98:
## Description

We have several PRs that contains table that overflows such as these examples:

- https://github.com/mautic/user-documentation/pull/841
- https://github.com/mautic/user-documentation/pull/279 (CSS fixed in this PR, but didn't touch the `make html` setting.

**The reason of overflow:**

Root-caused via local investigation: the CSS-only fix appeared to have no effect after rebuilding because `make html`'s incremental build reported "no targets are out of date" and skipped the static-file copy step entirely, leaving `docs/build/html/_static/tablefix.css` stale. This is a general gotcha (affects any future static asset edit, not just this one), so the build script fix is the durable solution rather than a one-off workaround.

### Changes

- `docs/css/tablefix.css:` added rules scoped to `.wy-table-responsive` table to fix table overflow:

  - `table-layout`: fixed so declared `:widths:` percentages are actually respected instead of being treated as hints under the browser's default `table-layout: auto`.
  - `word-break`: `break-word` / `overflow-wrap: break-word` on `code` inside table cells, overriding the theme's `white-space: nowrap` on inline literals so long unbroken tokens (no spaces) can break instead of overflowing.
- `.ddev/commands/web/build-docs`: changed `make html` to `make clean html` so every build is a full rebuild, guaranteeing static asset changes (like the CSS above) are always picked up — not just on `.rst` content changes.

### Trade-off

`ddev build-docs` now always does a full rebuild instead of an incremental one, so it will take a bit longer. Given the size of this docs site, this seemed like the right trade for correctness.

## Linked issue

N/A

### Before

Screenshot 2026-07-14 113113

### After

Screenshot 2026-07-14 113303


docs/css/tablefix.css
```diff
@@ -3,6 +3,16 @@
...
 
+.wy-table-responsive table {
+    table-layout: fixed;
+}
+
+.wy-table-responsive table td code,
+.wy-table-responsive table th code {
+    white-space: normal;
+    overflow-wrap: break-word;
+    overflow-wrap: anywhere;
```
adiati98: Remove double properties.

```suggestion
    overflow-wrap: anywhere;
```

escopecz (approved):
I see no issues. Thanks for taking care of this! 👍

PR #970, opened 2026-09-15
Document email address validation step for unsubscribe and resubscribe (8.0 backport)

promptless-for-oss:
## Description

Backport of the reviewed unsubscribe/resubscribe email-validation documentation (PR #913, branch `promptless/pr-16870-unsubscribe-validation`) onto the 8.0 branch, at maintainer @adiati98's request.

Mautic hardens the public unsubscribe and resubscribe flows against leaked Email links. The `{unsubscribe_url}`, `{unsubscribe_text}`, and `{resubscribe_url}` tokens resolve to a validation page: the Contact lands on a 'Confirm your email address' page and must enter the Email address the message was sent to before the unsubscribe or resubscribe completes. The Contact's Email address is no longer embedded in the first-click URL, which protects against leaked or forwarded links and blocks bot-triggered unsubscribes.

Updates the 'Unsubscribing' section of `docs/channels/emails.rst` (new confirmation workflow, mismatch error, legacy-link behavior, and bot-unsubscribe rationale), documents the new **Require email validation for unsubscribe** toggle in the 'Unsubscribe settings' section of `docs/configuration/settings.rst`, and adds a cross-reference from `docs/contacts/preference_center.rst`. The content stays aligned with the reviewed changes on PR #913.

**Latest change (link fix ported from #913)** — commit ports the link fix @adiati98 approved on #913 to this 8.0 backport, at the maintainer's request. In the **Require email validation for unsubscribe** bullet of `docs/configuration/settings.rst`, the closing sentence's whole-page reference `See the Unsubscribing section of the :doc:`/channels/emails` documentation …` becomes the section cross-reference `See the :ref:`Unsubscribing section of the Email documentation <Unsubscribing>` …`, so it links directly to the "Unsubscribing" section rather than the whole page. The `:ref:` target resolves to the "Unsubscribing" heading present in `docs/channels/emails.rst` on this branch. This is the identical link form already applied and approved on the 7.3 sibling PR #913; the two now read the same at that spot.

### Reviewer-feedback disposition (this pass)
- @adiati98 "apply the link fix to #970" (requested on #913): **applied verbatim** — the `:ref:` section-link form from #913's commit `72c87c36` was reproduced on `docs/configuration/settings.rst`. Verified the `:ref:` target `Unsubscribing` resolves unambiguously on this branch.

## Screenshots or screen recordings

Not applicable — RST documentation-only change (a cross-reference link form).

---
[...]

PR #972, opened 2026-09-15
Fix linkcheck timeout on stale HubSpot legacydocs link (8.0)

promptless-for-oss:
[Open in Promptless](https://app.gopromptless.ai/suggestions/0e6164eb-68be-4bcd-a9f5-68b6bdb9ae85)

The docs `build` CI job fails at the `make checklinks` step because the external link `https://legacydocs.hubspot.com/docs/methods/companies/get_companies_modified` (defined in `docs/links/hubspot_doc.py`, rendered at `docs/plugins/hubspot.rst` line 98) now 308-redirects to `developers.hubspot.com` and hangs past the link checker's 30-second timeout. The Sphinx HTML build itself succeeds; only the link check fails, and it fails on every PR to this repo regardless of the change.

This adds `https://legacydocs.hubspot.com/*` to `linkcheck_ignore` in `docs/conf.py`, following the repo's existing pattern for domains that block or time out the checker (Twilio, MaxMind, howtoforge, etc.). The link still works in a browser, so the reference itself is left in place.

Files touched: docs/conf.py.

**Trigger Events**
- [GitHub CI failure on user-documentation PR #970](https://github.com/mautic/user-documentation/actions/runs/34958721018/job/104346918275)


RCheesley:
@promptless-for-oss wouldn't it be better to update the link to the final destination, which seems to be https://developers.hubspot.com/docs/api-reference/legacy/crm/objects/companies/guide? Is there a reason why we should not do that which I am unaware of?

promptless-for-oss:
In reply to [this discussion](https://github.com/mautic/user-documentation/pull/972#issuecomment-5813804143)



Thanks for raising this — it's a fair question. Promptless is verifying where this HubSpot link should point, including whether the destination you suggested is stable, and will reply here with what it finds and any change to the fix.

promptless-for-oss:
In reply to [this discussion](https://github.com/mautic/user-documentation/pull/972#issuecomment-5813804143)



You're right, and there's no reason not to do it - I've updated PR #972 to point the docs link at its live final destination (https://developers.hubspot.com/docs/api-reference/legacy/crm/objects/companies/guide) and removed the temporary linkcheck-ignore entry. I verified that the old legacydocs.hubspot.com URL redirects (301, then 301, then 308) and lands exactly there, and that the final page returns HTTP 200 directly with no onward redirect in about 0.1s, so it won't re-trigger the link-checker timeout. This fixes the root cause instead of masking it, and since it was the repo's only legacydocs reference, dropping the ignore lets the checker keep watching the link for future rot.

Docs PR: https://github.com/mautic/user-documentation/pull/972

PR #973, opened 2026-09-15
Fix linkcheck failure on aivie.ch (WAF rejects the checker) (7.3)

promptless-for-oss:
[Open in Promptless](https://app.gopromptless.ai/suggestions/521bb7ed-4be6-461d-84fa-d11adc53b05e)

The Sphinx "Check links" (linkcheck) job fails on every user-documentation PR because of a pre-existing broken external link on the base branch: https://aivie.ch returns HTTP 415 to the linkcheck client. This link is referenced in `docs/builders/email_landing_page.rst` via the `:xref:`Aivie`` role (URL defined in `docs/links/alvie.py`), where it credits Aivie for making the GrapesJS-based Email and Landing Page builder available to the Mautic community.

The site itself is live and valid for readers — it returns HTTP 200 to browsers and to curl across every user-agent and request-method variant tested. The 415 is a checker-specific WAF rejection, the same "works in a browser but blocks the checker" pattern already handled for other domains in the `linkcheck_ignore` list (Twilio, linuxize, vtiger, howtoforge, HubSpot legacydocs, X developer).

Since the reference is correct and the destination works for users, this adds https://aivie.ch to the `linkcheck_ignore` list in `docs/conf.py` with an explanatory comment, rather than changing or removing the link. This unblocks the linkcheck job for all PRs while keeping the Aivie credit intact.


RCheesley (approved):

RCheesley:
Not sure if this one needs a back/forward port.

promptless-for-oss:
In reply to [this discussion](https://github.com/mautic/user-documentation/pull/973#issuecomment-5813817231)



Good question — Promptless is checking which maintained version branches carry this same aivie.ch link and would hit the same linkcheck failure. It will reply here with whether a back/forward port is needed and open the ports where it is.

promptless-for-oss:
In reply to [this discussion](https://github.com/mautic/user-documentation/pull/973#issuecomment-5813817231)



Yes - it needs both a back-port and a forward-port. The aivie.ch link and the linkcheck configuration are identical on every maintained branch, and the 'Check links' job runs on all of them, so the same HTTP 415 WAF rejection happens on 8.0, 7.2, 7.1, 7.0, 6.0, and 5.2 as well as 7.3. Promptless opened one PR per branch to add https://aivie.ch to linkcheck_ignore, so together with this 7.3 PR every maintained branch is covered.

8.0: https://github.com/mautic/user-documentation/pull/1000
7.2: https://github.com/mautic/user-documentation/pull/1002
7.1: https://github.com/mautic/user-documentation/pull/1003
7.0: https://github.com/mautic/user-documentation/pull/1004
6.0: https://github.com/mautic/user-documentation/pull/1005
5.2: https://github.com/mautic/user-documentation/pull/1001

PR #978, opened 2026-09-17
Warn that deleting a Point Trigger removes its event configuration and trigger history

promptless-for-oss:
[Open in Promptless](https://app.gopromptless.ai/suggestions/13721fd9-90d4-4282-849b-6a7eab57d07c)

## Description

Adds a warning to the Point Triggers section of the Points documentation explaining that deleting a Point Trigger permanently deletes its event configuration and trigger history, and that the action can't be undone.

The Points page previously documented creating Point Triggers but said nothing about deleting them or the consequences of doing so. Deleting a Point Trigger already cascades to remove its event configuration and its trigger history — this reflects existing Mautic behavior, surfaced to users in `mautic/mautic` PR #17442, which clarifies the deletion confirmation dialogs (single and bulk) with the same warning. This documentation change records that behavior so readers understand the irreversible consequence before deleting a Point Trigger.

## Screenshots or screen recordings

**Trigger Events**
- [mautic/mautic PR #17442: Clarify point trigger deletion consequences](https://github.com/mautic/mautic/pull/17442)


docs/points/points.rst
```diff
@@ -67,6 +67,10 @@ Once you have decided and entered those options, go to the **Events** tab. Here,
 .. image:: images/new_points_trigger_events.png
     :alt: Screenshot of New Points trigger events
 
+.. warning::
+
+   Deleting a Point Trigger permanently deletes its event configuration and trigger history. You can't undo this action.
```
promptless-for-oss, line 72: **Citation**



Deleting a Point Trigger cascades removal of its associated TriggerEvent records (the trigger's event configuration): the Trigger-to-TriggerEvent relationship is mapped with cascadeAll(), so Doctrine removes child TriggerEvent entities when the parent Trigger is deleted.

Source: [Trigger.php:150–156 · 44f7668](<https://github.com/mautic/mautic/blob/44f76686ae35bf76de4f09ce9083253a59bfd6af/app/bundles/PointBundle/Entity/Trigger.php#L150-L156>)

promptless-for-oss, line 72: **Citation**



The point_lead_event_log table (a trigger's history of firing events) has its event_id foreign key set to ON DELETE CASCADE, so removing the TriggerEvent rows (which happens when the parent Point Trigger is deleted) permanently removes the trigger's history rows as well, with no recovery path.

Source: [LeadTriggerLog.php:42–44 · 44f7668](<https://github.com/mautic/mautic/blob/44f76686ae35bf76de4f09ce9083253a59bfd6af/app/bundles/PointBundle/Entity/LeadTriggerLog.php#L42-L44>)

